PR #2, opened 2018-03-13
Import data sets

akva2:
Updated version of #1


akva2:
@alfbr i have one question regarding the original pr; should we really stick eclipse data in this repo?

atgeirr:
> should we really stick eclipse data in this repo?

If we intend to do some tests measuring "do we match eclipse within some tolerance" in addition to regression tests against current Flow results it would be needed. And we should add some such tests.

akva2:
i can't push here so my plan to merge stop here.

PR #12, opened 2018-05-14
manual update of the reference data for SPE5CASE1

andlaus:
this is necessary because of OPM/ewoms#319, but the underlying reason why no reference data was updated is most likely a bug in the Jenkins data update script.


andlaus:
jenkins build this please

PR #17, opened 2018-05-25
aquifer test case

nairr:
Adds a 2D OILWATER model containing 200 x 5 = 100 cells for testing aquifer functionality in FLOW


GitPaean:
And also, it can be a little more troublesome. Could change the file name to be something related to Aquifer. With the current file name, the testing message will be something like. 
```
60/68 Test #53: compareECLFiles_flow+2D_OILWATER ....................   Passed    0.72 sec
```
It might not be easy to tell it is the Aquifer model gets broken.

akva2:
why not use jenkins to generate the data as explained in the email ?

GitPaean:
> why not use jenkins to generate the data as explained in the email ?

Is it a general announcement? I do not recall it.

nairr:
@akva2  - Ah! is that what you meant by run update_data? I'm afraid I don't know how to use jenkins to do this

akva2:
you just include that in the trigger phrase from opm-simulators. let this pr only hold the setup, let jenkins add the reference data. i can help with those things, updating scripts and test cases etc if you tell me the pr chain required.

nairr:
the PR updating scripts compareECL.cmake and tests/update_reference_data.sh is **opm-simulators#1487**
and the PR with the aquifer implementation is **opm-simulators#1371**

PR #21, opened 2018-05-31
updating the reference results for PR 1478#

GitPaean:


GitPaean:
All the well curves were checked.

PR #25, opened 2018-06-20
Add wecon + wtest test

totto82:


totto82:
@alfbr This test should cover WTEST in combination with WECON. I plan to  add COMPLUMP to the test when the simulator is ready for it.  I takes 1 sec to run. I tried to make it as short as possible, but it need some time in order to first shut wells and then have time to re-open them again. 

This must be merged before I can test opm-simulators with updated reference data. i.e. I hope it can be merged asap.

PR #28, opened 2018-06-25
Added reference solutions for SPE1CASE1 using flow.

andreatlonn:
SPE1CASE1 gives correct answers using flow so the case is now supported by flow.


alfbr:
It is good to see more of the SPE-cases supported by flow, mering, thank you!

PR #33, opened 2018-06-28
Add COMPLUMP to 3D_WECON test

totto82:


alfbr:
This looks good, and highly appreciated! Please self merge when accompanying test of flow has gone through.

GitPaean:
Maybe it is good to rename the file name to be 3D_WECON_WTEST_COMPLUMP.DATA or something similar to distinguish this one to the 3D_WECON in opm-data. Different files with same name can always be confusing. 

And also, it can be good to indicate the testing feature in the file name.

alfbr:
That is a good point. I hope we can find time to think through and possibly reorganize the tests this summer. Filenames alone may not scale well when the numbers of tests grow significantly.

PR #34, opened 2018-06-29
add reference results for the new test for setting item 2 of EHYSTR to "1"

andlaus:
for some reason, they were not there.


andlaus:
since jenkins currently fails because of this, I'll emergency self-merge. if it causes trouble, please revert.

totto82:
@andlaus I reverted OPM/opm-simulators#1525. In order to make Jenkins happy. Sorry, I didn't know you where fixing it at the same time. I will not interfere anymore. Please revert, merge or do what is necessary to make Jenkins happy again.

andlaus:
with your opm-simulators revert I think that Jenkins is happy either way, so no bad feelings from my side. sorry for causing trouble...

PR #40, opened 2018-08-01
model1: do not write performance related summary fields

andlaus:
this is because doing so will cause any run to fail once these fields are actually supported by `flow`.

Note: I'm not 100% sure what ELAPSED is supposed to write. Wall time or simulation time? Anyway it does not make the test much less useful if that field is not written.


alfbr:
Agreed, we need different handling of the performance output if we are to test it. A better alternative to this patch would be to make compareSummary routine aware of the performance vectors, but this will take time to get into place. Please feel free to merge when you have new reference solutions ready and checked.

atgeirr:

andlaus:
jenkins build this please

PR #44, opened 2018-08-03
Changed the name of a file and added a test. 

andreatlonn:
Changed the name on BASE_MODEL_1 to STD_BASE_MODEL_1 and changed it so that it uses standard wells. Also included a test for multisegment wells similar to the original BASE_MODEL_1.


alfbr:
Looks good! Only two nits. Please rename STD_BASE_MODEL_1 back to BASE_MODEL_1, as the std and base prefix is redundant. Then rename MSW_BASE_MODEL_1.DATA to MSW_MODEL_1.DATA. For consistency you should then also rename the .sch files. I suggest history_basemod1.sch and history_msw.sch respectively. You also want to update the TITLE keyword within the MSW DATA-file to reflect the file name.

andreatlonn:
I know there is a conflict but I don't know how to solve it. I tried to solve it the same way as last time but it does not work since the file is not readable. This is one of the files that should be generated by update_reference_data.sh in opm-simulators/tests so maybe I can remove it from git and then it will be generated again by Jenkins?

alfbr:
I am on vacation, so a limited ability to check the conflict. From my end the pull request is good to go. @arnem can you do a final check and merge?

akva2:
i doubt he will ;)

for starters the PR has to be rebased.

PR #47, opened 2018-08-07
New case testing the FAULTS keyword.

andreatlonn:


alfbr:
Excellent! This is good to go from my end. @akva2 will you do a final check and merge?

PR #56, opened 2018-09-10
Two test cases for UDQ functionality added

tskille:
This is a starting point for test cases involving UDQ usage. Reference solutions have not been created, simply because the functionality is still not supported in flow.


alfbr:
This test-cases are very welcome! Please include header on the files. License, copyright and a brief explanation of the case on each .DATA file. Then these are good to go in.

alfbr:
Great, thanks!

PR #58, opened 2018-09-11
add SWATINIT to restart test

totto82:
There has been some issues when combining SWATINIT and restart. This PR adds a test for this by extending the SPE1CASE2_ACTNUM test. If we rather want to test this separately I can rename the file and add it as a new test.


alfbr:
From my end this looks good. Feel free to self-merge this with updated reference solution.

totto82:
I needed to re-add the EQUIL keyword with friends in order to make flow_legacy happy. (The legacy code will not read the SWATINIT scaling from the restart file, but re-generate it.

totto82:
We don't need any update of the reference solution here since the restart test compares the restart results with itself.

alfbr:
Ah, that is correct. Merging.

PR #65, opened 2018-10-16
Test case for editnnc keyword

tskille:
New test case for editnnc, to be supported by Flow. Grid with one fault which generates a number of nnc. The model is producing on one side of the fault. Block pressure 5,3,1 on the other side of fault is highly sensitive to nnc permissibilities.


alfbr:
Please check the copyright headers, they should probably all read:
`-- Copyright (C) 2018 Equinor`
Since this marks the start of a series of new keywords for integration tests, I suggest creating a new folder named model2, which can be a container for all keywords that needs to be tested together with EDITNNC.  Similarly as here:
https://github.com/OPM/opm-tests/tree/master/model1
With these two comments addressed, this is ready for merge.

blattms:
@tskille I have a question about the files. How are the NNCs that we edit actually defined? I could not find any explicit `NNC` definition. Might they be defined implicitly?
Could it be that an `EDITNNC` without any matching `NNC` defintion is not ignored in eclipse but treated
as there was an `NNC` definition with transmissibility 1 was there?

alfbr:
The nnc's are implicit by the faults. Connections between non-neighbors will then come after grid processing I believe. You can load the grid directly in resinsight (the input file), and see where cells connect over the fault without being ijk-neighbors.

alfbr:
Excellent, thank you! Merging.

PR #74, opened 2018-11-21
DRSDTR

btos80:
Split the model into 2 regions (pvt and relperm). Added keyword DRSDTR and gave different values for the two regions.


alfbr:
Thanks for the contribution! Both serves well as minimal integration tests for DRSDTR and AQUFET respectively. Merging!

PR #77, opened 2018-11-30
Test cases for model 2

tskille:
10 individual models which test majority of keywords used in model2.  0_BASE.DATA is a stripped version, 1_MULTREGT is 0_BASE + MULTREGT keyword, 2 MULTXYZ is 1_MULTREGT + MULX, MYLTY and MULTZ, and so on. 9_EDITNNC include all keywords used for model2


joakim-hove:
Fantastic

atgeirr:
That is great! Merge at will when complete.

alfbr:
Excellent, in it goes!

PR #87, opened 2019-01-10
adding reference result for 2D_FETKOVICHAQUIFER

GitPaean:
For the PR https://github.com/OPM/opm-simulators/pull/1642


alfbr:
Until we can automate it, manually works! Merging.

PR #108, opened 2019-04-05
Norne Porevolume Reference: Update for New Table Normalisation

bska:
The table normalisation (PR OPM/opm-common#707) changes the size and/or the contents of the
```
INTEHEAD, LOGIHEAD, DOUBHEAD, TABDIMS, TAB
```
keywords.

Additionally, this update also changes the reference data's `TRANNNC` vector.  The new update has 109 **more** zero transmissibilities across NNCs than the current reference data (2199 vs. 2090).  In the cases where both data sets have a positive transmissibility, these transmissibility values are bitwise identical in both the existing reference solution and this proposed update.

All other INIT file keywords&mdash;particularly `PORV`&mdash;are bitwise identicial in the original reference solution and in this proposed update.

Below is a plot of sorted NNC transmissibilities on linear connection index vs. logarithmic transmissibility value axes.  The cutoff point of 1.0e-6 suggests that the "missing" positive values can be attributed to the threshold that was introduced in commit OPM/ewoms@8f275478, PR OPM/ewoms#436 (OPM/opm-simulators@f9af2c50).

---

### Transmissibility Comparison Across Non-Neighbouring Connections ###

trannnc compare

---

### Update Following PR OPM/opm-simulators#1823 ###

Following PR OPM/opm-simulators#1823 (reworked version of PR OPM/opm-simulators#1818), we're no longer storing the zero NNC transmissibilities.  That reduces the size of the TRANNC vector from 15083 to 12884.  There is still the discrepancy of 109 NNC transmissibility values dropped due to PR OPM/ewoms#436 however.

trannnc compare


bska:
Merging now as part of OPM/opm-common#707.

PR #120, opened 2019-05-23
new test case for keyword WSEGVALV

tskille:
Test case (high viscosity oil) including one production well. Segment 8 - 14 has been defined with WSEGVALV keyword. Keyword WSEGVALV defines a segment to represent a sub-critical valve.

This case also include summary vectors/keywords SOFR, SWFR, SGFR, SPR, SPRD, SPRDF and  SPRDH. 

WSEGVALV keyword shall add an extra pressure drop to the friction element.


alfbr:
Perfect! Only comment is to consider grouping it with other test-cases later on (else we will get a very crowded opm-tests folder over time). Merging!

bska:
> This case also include summary vectors/keywords SOFR, SWFR, SGFR, SPR, SPRD, SPRDF and SPRDH.

:+1:

GitPaean:
> SPR, SPRD, SPRDF and SPRDH.

OPM has not implemented these keywords, as a result, OPM will not run this case because unknown keywords. 

Is it the intentional behavoir? I mean, stopping the simulation because of unknown summary keywords. 

@joakim-hove  @bska

bska:
> > SPR, SPRD, SPRDF and SPRDH.
> 
> OPM has not implemented these keywords, as a result, OPM will not run this case because unknown key

We are supposed to support [`SPR`](https://github.com/OPM/opm-common/blob/5f699679a9b48cd16be72939b1a8476ba0dd652d/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp#L533).  If we don't then that's a bug.  The others are unsupported as of right now.

> Is [stopping the simulation because of unknown summary keywords] the intentional behavoir?

I don't think so.  We're supposed to issue a warning about an unsupported summary keyword in that case and continue.  Where does the process stop?

GitPaean:
> We are supposed to support SPR.

You are right. It was because of other keywords. 

The following is the error message, 

```
   73 Reading SPR      in file /home/kaib/test/wsegvalv/include/segm_summary.inc, line 22

Error: Unknown keyword: SPRD
Failed to create valid EclipseState object.
Exception caught: PARSE_UNKNOWN_KEYWORD: Unknown keyword: SPRD
```

bska:
> The following is the error message,
> 
> ```
>    73 Reading SPR      in file /home/kaib/test/wsegvalv/include/segm_summary.inc, line 22
> 
> Error: Unknown keyword: SPRD
> ```

Ah, right.  In that case, I think you'll be able to get by that hurdle by adding those keywords to the [`SEGMENT_PROBE`](https://github.com/OPM/opm-common/blob/5f699679a9b48cd16be72939b1a8476ba0dd652d/src/opm/parser/eclipse/share/keywords/000_Eclipse100/S/SEGMENT_PROBE#L5-L10) specification.  Of course, actually calculating the pertinent values will require more work.

GitPaean:
> Ah, right. In that case, I think you'll be able to get by that hurdle by adding those keywords to the SEGMENT_PROBE specification. 

Sure. Thanks. I will put them in the WSEGVALV branch so it can run this specific case in opm-tests.

PR #122, opened 2019-06-04
test cases for wtest functionality and new prediction cases

tskille:
Two new prediction cases based on NORNE 2013 model. None of these cases have group control. First case have a minimum bottom hole pressure (bhp) constraint and second case have minimum tubing head pressure (thp) constrain. Case with minimum thp and vfp lift curves is not running with flow.

8 new test cases with WTEST keyword (periodic testing of closed wells). The WTEST keyword is currently not implemented correctly in flow. These test cases can be used for updating the WTEST implementation in flow.


atgeirr:
Very good to have this! Just one question: what is the op_1.ecl file? It is about 90% of the data in this PR, but github will not show its content to me because it is too large...

tskille:
op_1.ecl is a vfp lift curve. 3.1 Mb ascii file.

atgeirr:
> op_1.ecl is a vfp lift curve. 3.1 Mb ascii file.

What a beast!

Anyway, I am fine with merging this, unless you want to add more stuff to this PR?

tskille:
Yes, I believe this is ready to be merged. The plan is to add more cases later on, but for now it is good if this can be merged.

PR #126, opened 2019-06-09
Prediction cases based on Norne 2013 model.

tskille:
1A: Depletion case, wells controlled with maximum rates and minimum bottom hole pressure. No group controlls
1B: Depletion case, wells controlled with maximum rates and minimum tubing head pressure. No group controlls

remaining cases: wells constrained with maximum rates and minimum tubing head pressure. Ctrl Mode Group

2A: Depletion case. GCONPROD with constraints on one group level
2B: Depletion case. GCONPROD with constraints on two group levels

remaining of cases using GONPROD with constraints on one group level

3A: GasInj case. One gas injector. GCONINJE with reinjection (100%) og produced gas
3B: Same as 3A except adding gass export (GCONSALE from 1 october 1999)
3C: Same as 3B except adding gass consumption (GCONSUMP)

4A: WatInj case. Water injectors controlled with maximum BHP and maximum water injection rate. No group control on injection
4B: Same as 4A except GCONINJE with 100% reservoir volume replacement
4C: Same as 4B except adding gas reinjection (100%). Water injection is top-up.
4D: Same as 4C except adding gas export (GCONSALE) and increasing gas production capasity from Oct 2001
4E: Adding GUIDERATE kewyword set up to prioritize wells with lower GOR, maximizing oil production.

This is still work in progress. The plan is to use these models for implementing prediction functionalities in OPM flow. Later on, model2 will be used for making test models to be used for regression testing. This model is close to what would be a realistic prediction setup, but runtime prevent us from using this for regression testing.


bska:
jenkins build this please

bska:
I'll merge this into master now. It is a pure extension that expands our feature coverage and that does not alter any existing tests.  The failing tests are caused by unrelated issues.

PR #131, opened 2019-06-21
Terminate AQUCT With Empty Record

bska:
The specification states that the keyword should be terminated by an empty record containing just a slash.

ECLIPSE does appear to [accept](https://github.com/OPM/opm-tests/tree/5c943189238a16102adacef564abbc0ba0cdc35f/flow_diagnostic_test/eclipse-simulation) the original version though.


bska:
jenkins build this please

joakim-hove:
> An input file does not need to be executable.

I would even say it stronger: "An input file *should* not be executable"

PR #132, opened 2019-06-22
model2 - updated schedule section  for all cases 0_BASE to 9_EDITNNC

tskille:
Keyword WRFTPLT added to schedule section file used by all 10 decks. 

WRFTPLT is used to request simulation/modelling of RFT's for a selected well. Simulation results for these RFTs are stored in the RFT file. RFTs are used in the majority of history matching studies and these are very important data.

### DIFF
--- a/model2/include/hist_schedule.sch
+++ b/model2/include/hist_schedule.sch
@@ -7,6 +7,8 @@
 
 -- This file is one of the include files for model2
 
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

atgeirr:
I know there should be no need to test this with Jenkins, but what better way do check that assumption than using Jenkins...

Anyway, merged now.

PR #133, opened 2019-06-24
Update udq dataset

joakim-hove:
Make sure all UDQ data is output in the summary files.


joakim-hove:
jenkins build this please

PR #139, opened 2019-07-04
Add data file for small foam test case.

atgeirr:
The test case is based on SPE1CASE2 and is completely artificial. No attempts have been made to make it realistic at this point.


alfbr:
Consider including Sintef in the copyright header.

alfbr:
There are no comments regarding the foam related keywords in the file. If you intend this to be a regression test, then please include some explanatory lines. Preferably a couple of lines in the header explaining what you test, and then a few lines for the keywords with a brief explanation. Otherwise this looks great, thanks for taking the time to create the data set!

atgeirr:
Comments added and force-pushed.

alfbr:
Excellent, in it goes!

PR #161, opened 2019-09-06
fixed: disable TCPU and TCPUDAY

akva2:
these will be different in different runs,
leading to files always being updated when update_data is invoked.

ref https://github.com/OPM/opm-common/pull/996

@joakim-hove 

thanks to @bska for hand holding - i know more about these file formats now.

### DIFF
--- a/udq_actionx/include/example_summary.txt
+++ b/udq_actionx/include/example_summary.txt
@@ -145,8 +145,4 @@ WMCTL
 /
 WLPR
 /
---
--- Simulatior performance
-TCPU
-TCPUDAY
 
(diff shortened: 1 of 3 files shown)


akva2:

joakim-hove:
That makes sense - very good. I would just rip the offending keywords all the way out - it is not as if they will be coming back in the regression test one day.

joakim-hove:

joakim-hove:
Look like the offending keywords are here: `wtest/include/example_summary.txt` as well

akva2:
not used for testing though but i will nuke em there as well.

akva2:

akva2:

akva2:

akva2:
meh, update_data from here is not too common and i have to change things to facilitate it. i'll merge this and use another pr for the update.

PR #176, opened 2019-10-24
updated norne case 1a and 1b with multi-segment well model

tskille:
updated NORNE_AWT2013 prediction cases 1a and 1b with multi segment well model. 

STDW and MSW cases have identical completions (COMPDAT). 

STDW = standard well model, MSW = multi-segment well model


atgeirr:
Great, thanks for this case!

PR #188, opened 2019-11-25
Add test case with ROCKCOMP

totto82:


alfbr:
Thanks for the contribution, this is welcome! Please consider a pull request adding this as a regression test.

PR #189, opened 2019-11-25
new test case for WSEGSICD

tskille:


alfbr:
Excellent, test case is very welcome and all headers in shape. Merging.

PR #191, opened 2019-11-25
Add  SPE1CASE2_RADIAL output files from Eclipse

goncalvesmachadoc:


alfbr:
The jenkins server can generate reference solutions, so there is no need to include them. I believe we also will not maintain Eclipse solutions moving forward (they only make sense when there is a marked difference between Eclipse an Flow). Let us keep this open until we have are ready to put this up as a regression test, then we can close it.

PR #193, opened 2019-11-26
added: add well curves for parallel simulation

akva2:


akva2:
this should be uncontroversial, proved working and is testing related so kinda my turf. shameless self-merge.

blattms:
Cool, thanks.

blattms:
But the parallel is not run by default, right? Hence this will only help if the previously merged PR did request a parallel run (and the wells were checked). Otherwise the bug might have sneaked in earlier.

akva2:
yes for the PR builder you will have to run both. It's mainly for the post builder where both are always run.

PR #213, opened 2019-12-18
MULT Tests

OPMUSER:
New set of MULT test cases and documentation. See EIPC OPM Flow MULT Runs Rev-2.pdf file for documtation.


blattms:
I am wondering whether we really want binary files in a repository. How much space does the zipping save.

OPMUSER:
It is not so much the saving of space, but making the directory more organized. So that the cases that have been run with various version are easily identified.

bska:
I like this version a lot and I think it should be merged into master.

alfbr:
Excellent, I agree. In it goes :-)

PR #214, opened 2019-12-18
test WELTARG and WEFAC

totto82:
I suggest the following changes to cover WEFAC and WELTARG


totto82:
Note this will need OPM/opm-simulators#2235 and OPM/opm-common#1351 to fly

akva2:
this will not affect any tests - there are no tests using SPE9_CP.DATA, only SPE9_CP_SHORT and SPE9_CP_GROUP

totto82:
Good point. I will add the same modifications to SPE9_CP_SHORT instead. Thanks for noticing!

totto82:
I have made a new PR for SPE9_CP_SHORT

PR #226, opened 2020-01-03
Remove Empty RFT Regression Test Output Files

bska:
These were added as a result of a bug in the RFT file-writing code that ended up creating RFT files even if no RFT output was requested.

The latter problem is corrected in OPM/opm-common#1387, so this PR must not be merged until that is installed in opm-common's master branch.


bska:
jenkins build this please

bska:
Merging this now to prepare merging OPM/opm-common#1387

PR #231, opened 2020-01-29
Update sp9 restart

joakim-hove:
This is a preparation step for: https://github.com/OPM/opm-common/issues/1396

As part of https://github.com/OPM/opm-common/issues/1396 there will be a stronger coupling between the input deck and the restart file - in the case of a restarted case the restart file can semantically be perceived as an include file.

The first commit makes sure that the date we are restarting from is explicitly mentioned in the Schedule section - this is *highly recommended* in the manual. The second commit makes the restart file available (as a symlink) from the location of the input file.


joakim-hove:
jenkins build this please

PR #233, opened 2020-01-30
Add second PLMIX region

totto82:
Needs fix in OPM/opm-models#588


bska:
@totto82 : Is this ready for merging now?

totto82:
yes. he values for the regions are the same i.e. the results of the test case should not change i.e. we can merge this without any update of data.

bska:
> the results of the test case should not change i.e. we can merge this without any update of data.

Cool.  Do you want to run build check just in case?

totto82:
Can I do that from opm-tests?

bska:
> Can I do that from opm-tests?

Yes.  We can run "build this please" from opm-tests, but we can't trigger reference data updates from this repository.

totto82:
jenkins build this please

bska:
Build check was successful.  I'll merge into master.

PR #234, opened 2020-01-31
update case: wecon+max_wgr, strong aquifer support

tskille:
Fixed test case for WTEST + WECON (MAX_WGR). 

Current test case shut down well due to physical conditions before reaching max water gas ratio set in WECON.


GitPaean:
Thanks a lot.  Please let us know when it is ready to merge from your side.  There is currently no test case using this deck, so there is no need to update reference result.

GitPaean:
jenkins build this please

PR #235, opened 2020-02-02
Ensure that comment starts in column 9

joakim-hove:
The dataset had comments starting in column 8 - that could create problems


joakim-hove:
jenkins build this please

PR #240, opened 2020-02-17
Add DATES keyword for restart date in testdata

joakim-hove:
When restarting with the `SKIPREST` keyword you *really* should have the restart date in the deck.


joakim-hove:
jenkins build this please

PR #241, opened 2020-02-19
Add testcase for restarting NORNE

joakim-hove:
I *think* [this issue](https://github.com/OPM/opm-simulators/issues/2354) shows that we are currently unable to restart norne. This PR consists of three files:

1. The file `NORNE_ATW2013_RESTART.DATA` is a normal norne deck where the initialization has been replaced with the `RESTART` keyword, the `SKIPREST` keyword has been added to the `SCHEDULE` section and a shortened schedule file is used.

2. The schedule file `BC0407_RESTART.SCH` is a shorthened version of the schedule file which only contains one `TSTEP 2`after the restart date.

3. The file `NORNE_ATW2013.UNRST` is the restart file we should start from - it only contains one report step.

To actually *use* this will require a designated run script in opm-simulators; to start with my plan was something which just invoked and initialized flow - and examined exit status?


akva2:
that's precisely what the 'compareECLInitFiles' does so maybe that can be adapted.

joakim-hove:
> that's precisely what the 'compareECLInitFiles' does so maybe that can be adapted.

Sounds like a plan.

joakim-hove:
Ping!

This has not received much attention:

1. Is it interesting - do we want this? 
2. In  the [original issue](https://github.com/OPM/opm-simulators/issues/2354) the 19.10 release works, whereas this along with https://github.com/OPM/opm-simulators/pull/2358 fails hard. So unless I have done something wrong in the current testcase the norne restart is sometime between the 19.10 release and now. Would be cool if someone else would test this.

NB: This is 100% independent of the ongoing restart without historical schedule file work.

OPMUSER:
I cannot find the files to test with 1910 to confirm we have a problem, can you point me in the right direction?

joakim-hove:
> I cannot find the files to test with 1910 to confirm we have a problem, can you point me in the right direction?

The files are part of this PR - fasten seat belts and prepare for some git joy:
```
cd opm-tests
git remote add joakim git@github.com:joakim-hove/opm-tests
git fetch joakim
git checkout joakim/norne-restart
```
If you now do `cd norne` you should have a new file `NORNE_ATW2013_RESTART.DATA`- with the accompanying Schedule file and restart file in `restart/NORNE_ATW2013.UNRST`

joakim-hove:
After private discussions I will now merge this; actually fixing the problem is a next step ...

PR #251, opened 2020-03-04
fixed: add FOAMOPTS to SPE1FOAM case

akva2:
This was missing from this case.


bska:
I thought you were allowed to have foam-related data without specifying FOAMOPTS.  Is that not the case?

akva2:
well, yes, you did. it's a fallout from my change in FoamConfig, https://github.com/OPM/opm-common/commit/26bd5fd4a0ee0b313ffd6dc6e52c8cbea2392995

the mobilitymodel defaults to INVALID. when i start using the values from the config in https://github.com/OPM/opm-models/pull/600/commits/e2476dea3e706319608eda1fa361a305b84c3006 it requires them (strictly speaking only the mobility model) to be specified.

the alternative fix is to default to TAB in FoamConfig. i figured this was better as it makes things explicit. i did not realize this before continuing the work..

atgeirr:
Since we must allow FOAMOPTS to be defaulted, this deck change should not be necessary. We should rather change the code.

Default phase should be GAS, default mobility model depends on phase: TAB for GAS, FUNC for WATER.

akva2:
okay, so i change it to TAB since that is what is supported now. thanks.

PR #265, opened 2020-03-17
updated test model for wsegvalv

tskille:
The previous version of this model had issues (which is not resolved) not related to wsegvalv. This version of the model is a simplified version of the previous version. 

This model still have significant impact of the inflow device (keyword WSEGVALV).


GitPaean:
Thanks @tskille for updating the case. 

I wonder whether in the old version, there are some keywords are not well supported yet.

PR #274, opened 2020-03-31
update of model4

tskille:
* redused memory requirements, relevant for Eclipse simulator
 * modified EQUIL records item9


atgeirr:
Thanks, merging!

PR #292, opened 2020-04-20
Added case with fault multipliers in schedule section.

alfbr:
Added one case based on the model2 case to test fault multipliers in the SCHEDULE section.

### DIFF
--- a/model2/3_A_MPI_MULTFLT_SCHED_MODEL2.DATA
+++ b/model2/3_A_MPI_MULTFLT_SCHED_MODEL2.DATA
@@ -1,0 +1,236 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2018 Equinor
+
+-- Testing FAULTS and MULTFLT keyword ni combination with MPI
+-- Here the fault multipliers have been applied at 1 JAN 2020
+-- i.e., in the SCHEDULE section.
+
+-- ------------------------------------------------------------------------
+-- ----------------------- BASE MODEL 1 -----------------------------------
+-- ------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


joakim-hove:
More tests - I am all for it; but as a general comment I think the new tests should be added with an accompanying PR in opm-simulators which actually uses the test; otherwise we risk that the test just hangs around int this repo without actually being used.

akva2:
it will come. i just have to fix the bug first.

alfbr:
>  I think the new tests should be added with an accompanying PR in opm-simulators which actually uses the test
In principle I agree. However, from a practical perspective this is a multistage process. We only add tests in opm-simulators once the results have been checked to be good. Moreover, the tests are also used to expose new feature needs, hence putting them into opm-simulators is not possible until the feature is implemented, e.g., the UDQ/ACTIONX stuff.

In this PR, the test-case exposes a bug and is needed for the bug hunting. Once the bug is tracked, and the solution is confirmed good, it will be added. Having pull requests open for prolonged periods is not practical.
> we risk that the test just hangs around int this repo without actually being used
I do not see this as a problem actually.

bska:
jenkins build this please

alfbr:
Any reason for not merging?

bska:
> Any reason for not merging?

Absolutely not.  I think this should be in master.  The only thing holding me back was that @akva2 mentioned he was working on a regression test/bug fix based on this dataset and I didn't know if he wanted to complete that work before this was merged.

As far as I'm concerned, this should go into master right now.

alfbr:
> The only thing holding me back was that @akva2 mentioned he was working on a regression test/bug fix based on this dataset and I didn't know if he wanted to complete that work before this was merged.

I see. Yes, the bug-fix has confirmed by me and merged to master. The PR is found here: https://github.com/OPM/opm-common/pull/1721

bska:
>> The only thing holding me back was that @akva2 mentioned he was working on a regression test/bug fix based on this dataset and I didn't know if he wanted to complete that work before this was merged.
> 
> I see. Yes, the bug-fix has confirmed by me and merged to master. The PR is found here: OPM/opm-common#1721

Very good.  Thanks for the heads up!  I'll merge this into master now.

PR #293, opened 2020-04-20
Reduced time step size and runtime to improve accuracy.

alfbr:
This is the first step in making this test pass locally for users. The runtime has been significantly reduced while maximum timestep has been significantly reduced. This may be sufficient to make tests pass locally.


alfbr:

alfbr:
I suggest this is merged once an updated reference solution is available. While we still do not know if this fixes the local failures, it presents a much better starting point, and still tests the functionality just as well as before.

bska:

@akva2 : Does Jenkins support updating reference solutions directly from an opm-tests PR?

akva2:
No.

alfbr:
Any suggestions on how to proceed @akva2 ? Should I open up a dummy PR in opm-simulators?

akva2:
the problem is we can't have two opm-tests PRs in a single build. so you won't be able to check the generated data even if you do that. so only way to go about it is a dummy PR in opm-simulators, and then you have to merge the two opm-tests PR's manually (this and the generated data).

alfbr:
I have a couple of more PRs in motion, I suggest we just use one of them. There is no hurry.

alfbr:
I have a well-suited candidate to use to get updated solutions for this case. It is found here https://github.com/OPM/opm-simulators/pull/2553

PR #297, opened 2020-04-23
Test case including network option and GL optimization

tskille:
This is a first version of a new suite of models. There will be more models added. 

- 0_BASE_MODEL5.DATA - basic model with group control
- 1_NETWORK_MODEL5.DATA - based on above but extended with network
- 2_GLIFT_MODEL5.DATA - based on above but extented with gass lift optimization


alfbr:
Excellent, unless somebody objects, this is going in. Maybe you want to update the year in all headers to reflect that these were made in 2020 and not 2018?

tskille:
headers updated 2018 -> 2020

bska:
> Excellent, unless somebody objects, this is going in

👍

PR #295, opened 2020-04-21
update reference data for 0a1_grpctl_msw_model2 0a2_grpctl_msw_model2 0a3_grpctl_msw_model2 0a4_grpctl_msw_model2

jenkins4opm:
Reason: https://github.com/OPM/opm-simulators/pull/2552

opm-common     = 3c60cb3765dff2dfd5d4068e45c33dd2467f5726
opm-grid       = e2511eda3d4ff93a49af968a940e26bbf7bc5348
opm-material   = afec03758cb7d3d61511ed16c663de4976b9b22b
opm-models     = d924d4fea8382df1677fb375847e58c9a0e993fa
opm-simulators = 62d3f87100b8e2f1d99108b92fcdd199fd04f43f

PR #299, opened 2020-04-24
Combined PRs 293 and 295

atgeirr:
Manually created since Jenkins cannot simultaneously use and update opm-tests PRs.


atgeirr:
jenkins build this please

alfbr:
Are there still issues we need to resolve?

bska:
> Are there still issues we need to resolve?

I don't think there are any issues remaining from the implementation side.  There appears to be some conflicts in a couple of `.UNSMRY` files in this PR which may need manual intervention.

@akva2 : Any suggestions on the appropriate course of action for resolving this, #293, and #295, in addition to the OPM-Simulator PRs OPM/opm-simulators#2552 and OPM/opm-simulators#2553?

akva2:
if things are dependent, they need to go into the same PR's. there is no way around it, we can only build with one PR per repo.

alfbr:
This got too complicated. I suggest we take one PR at the time, and start with https://github.com/OPM/opm-simulators/pull/2553

bska:
> if things are dependent, they need to go into the same PR's.

Okay, I don't think that's actually feasible here.  Instead I think we'll have to do a staggered sequence of commits.  I'd suggest the following sequence:

1. Close this and #295 and merge #293 as is.  This will possibly introduce a temporary test failures.

2. Update reference data from OPM/opm-simulators#2553 and commit the results along with the opm-simulators PR.

3. Then decide whether we still want/need OPM/opm-simulators#2552 and, if so, launch another data update from that PR and commit the result.

alfbr:
Agreed. Closing this one now. Having https://github.com/OPM/opm-simulators/pull/2552 here was an error that probably stemmed from me mixing up PRs in opm-tests. That one will need some more iterations before it is ready for merge.

PR #309, opened 2020-04-29
Two extra test cases for gas export and fixed error in 9_2B

tskille:
Two extra test cases which is targeted to test item3 and 4 in GCONSALE. 
Fixed one mistake in in existing test case   

- 9_2B_DEPL_GCONPROD_2L_MSW.DATA  
- 9_2B_DEPL_GCONPROD_2L_STW.DATA


bska:
jenkins build this please

bska:
I'm preparing to merge OPM/opm-common#1749 which will also require updating the reference data. I'll merge this now, temporarily creating test failures, and then update everything as part of merging the other PR.

alfbr:
> I'll merge this now, temporarily creating test failures

It should not, none of these cases are tested yet.

bska:
>> I'll merge this now, temporarily creating test failures
> 
> It should not, none of these cases are tested yet.

Right, but this PR requests a couple of new summary vectors in an include file that is common to many other, existing regression cases.  That's what causing the test failures.

alfbr:
Ah, I see, very impressed you picked up on that :-)

PR #319, opened 2020-05-07
fixing input error in both MOD4 models

tskille:
Error in data deck fixed. In current version properties for 'WID*' wells is set **before** 'WI*' which means (at least in Eclipse) that the first record have no effect 
```
WCONINJE
 'WID*' 'WATER'   'SHUT'  'RATE'    2000        1*      210. /
 'WI*'  'WATER'   'SHUT'  'GRUP'    2000        1*      210. / 
/
```


bska:
> In current version properties for 'WID*' wells is set before 'WI*' which means (at least in Eclipse) that the first record have no effect

Oooohh&mdash;spooky.  I don't know how Flow behaves here&mdash;summoning @joakim-hove to elaborate&mdash;but I can at least launch a test run.

bska:
Jenkins build this please

joakim-hove:
> In current version properties for 'WID*' wells is set before 'WI*' which means (at least in Eclipse) that the first record have no effect

Ahhh - that was a new twist; here we will do the same as Eclipse - which I actually think is quite logical, although certainly a bit scary.

tskille:
Just to be clear. This was a user error from my side when I created the test deck.  I have checked and we do the same as eclipse with respect to parsing and processing of the WCONINJE records. 

My intentions when creating the test deck, was to set water injectors WID1 and WID2 on rate control, but the second record did mess that up.

alfbr:
Thanks for the fix. The failing test is the expected one. Hence, I suggest we update the data and merge this.

alfbr:

bska:
> ```
> ```

Our Jenkins system does not support generating new reference data from a PR in opm-tests.  We need to merge this and then use a PR in one of the other repositories to get new reference solutions.  I sort of have a candidate PR in OPM/opm-common#1770, but I'm awaiting testing.  PR OPM/opm-simulators#2600 might be a better vehicle.

alfbr:
Thanks for the reminder! Still a change has been submitted since last build. Let us trigger a new build before merging.

alfbr:
jenkins build this please

alfbr:
Build finished with the one expected failure. Since this needs an update data we need to coordinate this with a PR. It seems the suggested one is already merged. Maybe we can use https://github.com/OPM/opm-common/pull/1784 ?

bska:
> Maybe we can use OPM/opm-common#1784 [as a way of getting new reference data]?

Yes, that's a good idea.  That PR seems close to finish.  We just need to ping the reviewer to get an all-clear.

bska:
I'm merging this now and then I'll use OPM/opm-simulators#2600 as a way to generate updated reference data.

PR #327, opened 2020-06-07
Add restart case with UDQ / UDA values

joakim-hove:


alfbr:
I am a bit in the dark, my apologies in advance. Is this deck identical to the one it is supposed to restart?

alfbr:
I was wrong, sorry for the noise. Thanks for the contribution, we really need to get restart tests going for this functionality. In it goes.

PR #331, opened 2020-06-12
Updated mode1 with 3 new cases for segment pressure drop calculations

tskille:
These 3 models are set up with segment pressure drop calculations which includes 

- hydrostatic
- friction
- acceleration  

Summary section is set up with SPRD, SPRDH, SPRDF and SPRDA.


alfbr:
Excellent, these data sets are very valuable for the multi-segment well model. Merging.

PR #337, opened 2020-06-15
One extra test case for model4.

tskille:
includes extra ACTIONX - 02,  which shuts water injector with lowest rate when total number of injector exceeds 6. Total number of injectors in the model is 8. Hence this action will be triggered twice.


alfbr:
Excellent, good coverage of the ACTIONX functionality is a necessity. Merging.

PR #343, opened 2020-06-25
Test case for WATVISCT

goncalvesmachadoc:
Create test case for water dependency on temperature. This case run with with fix in PR OPM/opm-material#422..


atgeirr:
Thanks, I'll merge so this can be made into a regression test case.

akva2:
forgotten button?

atgeirr:
> forgotten button?

Yes. I am happy that GitHub does not try to anticipate me and merge things it thinks I may want to merge though...

PR #349, opened 2020-08-11
3 new test cases with full field examples of UDQ 

tskille:
One real ffm example of UDQ. Udq simulation results output to summary file.


bska:
jenkins build this please

joakim-hove:
Can this be merged?

tskille:
It is ready from my end.

alfbr:
Only two nits:
-There are three models, with identical problem description in the header. If at all possible, include one more sentence so that it will be clear what is tested for when the test breaks.
-The grid file is missing license/copyright header.
Then this is ready for merge.

tskille:
license/copyright header added for grid include file.  

Problem description in the header for  UDQ_M1.DATA  and UDQ_M2.DATA are still identical. These are two real examples being used in a FFM.  I suggest that may be @joakim-hove can update these headers when the udq functionality is implemented and before we put these models into regression testing.  It may be that we don't need to use both of these for regression testing.

alfbr:
@joakim-hove do you have a suggestive descrtiption for M1 and M2? I really would like to have at least a brief description of what the data set tests before merging.

joakim-hove:
> @joakim-hove do you have a suggestive descrtiption for M1 and M2?

I have added a PR to this PR here: https://github.com/tskille/opm-tests/pulls - with a brief description of what is tested from my perspective

alfbr:
Excellent, in it goes!

PR #352, opened 2020-08-20
Mod4 restart

joakim-hove:
Add restart case for `MOD4_GRP`

Test actually using this here: https://github.com/OPM/opm-simulators/pull/2729


joakim-hove:
This is not in use; I'll merge for simpler development/testing of restart.

PR #360, opened 2020-09-07
Adding one more test case which th PINCH ITEM4=ALL

tskille:
* introduces a vertical flow barriere with MULTZ=0.0 for a pinched grid cell

This case is equal to case PINCH_MULTZ_ALL except for the addition 

```
EQUALS
 'MULTZ' 0.0 4* 10 10 /
/
```

Layer 10 is collapsed (hence inactive), but when used together with PINCH (item5 ALL) this should act as a barrier for vertical flow.  Current version of Flow is not honoring this flow barrier


alfbr:
Excellent, in it goes!

PR #366, opened 2020-09-08
Request FOPR output

joakim-hove:
The OPM UDQ code requires that summary variables used in the UDQ evaluation are also requested as summary output.


alfbr:
This sounds like working around a short coming in Flow. If a summary vector is used in a UDQ, it should be calculated by Flow without needing to be specified in the summary section? Only trying to understand the issue at hand at this point.

joakim-hove:
> This sounds like working around a short coming in Flow. If a summary vector is used in a UDQ, it should be calculated by Flow without needing to be specified in the summary section? 

Yes - you are right. We should create functionality to scan the UDQ section and ensure that all required summary vectors are calculated; it is a semi large job and for now I suggest just we just add `FOPR` to deck as in this PR.

alfbr:
I leave that decision up to you. If you see it as beneficial to activate the test before all the functionality is in place, go ahead and merge the change.

joakim-hove:
OK - I'll merge the current approach - I have even added a couple of more keywords with the same reasoning. I'll think about the superior "scanner" solution and get back to it.

PR #372, opened 2020-09-14
New 2 phase model (oil + water). Two data files

tskille:
* standard well model
* multi segment well model


alfbr:
Thanks, perfect! Merging.

PR #375, opened 2020-09-14
Add ref for model 6 1_MSW_MODEL

totto82:


alfbr:
Besides the just merged name change, please note that we do not track DBG or PRT files. In addition, we let jenkins generate reference solutions.

PR #376, opened 2020-09-14
add test case model 6

totto82:
Now with updated case name.


bska:
> Is there a reason for not letting jenkins provide the reference solution?

I'll echo @alfbr here.  Why do we not generate reference solutions directly on the Jenkins CI system?

alfbr:
> > Is there a reason for not letting jenkins provide the reference solution?
> 
> I'll echo @alfbr here. Why do we not generate reference solutions directly on the Jenkins CI system?

Thinking about it, it will be interesting to see if Jenkins agrees with the reference solution. We are having problems reproducing the solutions for Jenknins on tests with multi-segment wells. Hence, there is at least one reason :-)

bska:
> it will be interesting to see if Jenkins agrees with the reference solution.

It [does not](https://ci.opm-project.org/job/opm-simulators-PR-builder/1253/testReport/), so that's at least one more datapoint.

totto82:
I will close this and let jenkins compute the ref. solution.

PR #380, opened 2020-09-21
Re-perf well, don't update ref depth

tskille:
Test based on NORNE_ATW2013. This model included only one well (B-1H) which original sequence of perforation and then additional perforation. 

WELSPECS is set up with item5 (reference depth) defaulted, which means that this depth is defined from cell center depth of first connection according to COMPORD. 

Notice that reference depth **should not** be reset (updated) when additional perforations (in this case shallower) are added.   

Notice that bottom hole pressure (WBHP:*) is reported at reference depth.


tskille:
notice that if the WELSPECS keyword has been added also at  28 'MAR' 1999 (date for the additional perforations), then reference depth must be updated. If only COMPDAT keyword, then this should not be updated.

alfbr:
Perfect, thank you!

PR #383, opened 2020-09-21
test case combining network modelling and MSW

tskille:
- re-naming original case 1_NETWORK_MODEL5 with _STDW
- adding one more case that combines network option with multi-segment well model


atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

PR #385, opened 2020-09-23
Test case with analytic aquifer - CT.

tskille:
Based on 0_BASE_MODEL6
 - using keyword AQUCT with item3 defaulted. This is not working with current version of Flow. 
 - This should set initial pressure for the aquifer so that the aquifer is in equilibrium with the reservoir


GitPaean:
The initial pressure part is the same bug fixed with Fetkovich aquifer.  Will create a PR for initial pressure part soon.

alfbr:
Excellent, thank you!

PR #393, opened 2020-09-28
fixing inconsistency in data deck

tskille:
- cell 20 1 10 and 20 2 10 is used to model aquifer 1 and 2
 - aquifers no longer connect to these cells


alfbr:
jenkins build this please

PR #396, opened 2020-09-29
Added small test with MIN/MAXVALUE for parallel testing

blattms:
The test based on model5 failed in parallel in parts not connected to transmissibility. The new test use model2 as a base ot prevent this.


bska:
@blattms : What is the state of this work?

blattms:
Closing this one as stale

PR #399, opened 2020-10-07
A radial model test case for the STONE2 three-phase relative permeability model

jalvestad:
This model is a corner point grid approximating a radial sector model
The fluid is a gas-condensate system and there is one well in the innermost cell in the model
The reservoir is produced by the single well using pressure depletion

The model is a test case for  the STONE2 three phase relative permeability model
The model develops a three phase system in the near-well region in the model that influences the pressure loss and production rate for the well


alfbr:
Excellent. Please add copyright and license header on all include-files, then we merge.

jalvestad:
I have added copyright statements to all files as requested.

bska:
> I have added copyright statements to all files as requested.

Thanks.  I'll merge this now.

PR #424, opened 2020-10-28
Request WOPRL in test dataset

joakim-hove:
Adding completion summary keyword WOPRL to testcase. This is to support the in-progress work of creating summary output for completions. This is *not* ready to be merged yet.


bska:
> This is _not_ ready to be merged yet.

@joakim-hove : What's the state of this work?

joakim-hove:
We support completion output in the summary files now - the status of this particalr PR I don even remember. Will close it.

PR #427, opened 2020-11-04
Completion summary vectors

tskille:
New test case for regression testing of completion summary vectors.

WGIRL, WGITL, WWIRL, WWITL, WOPRL, WOPTL, WGPRL, WGPTL, WWPRL, 
WWPTL, WLPTL, WWCTL and WGORL


alfbr:
Excellent, please update the case description with the suggested expansion and in it goes.

PR #433, opened 2020-11-11
Test case for numerical aquifers

tskille:
- 3 aquifers cells with same aquifer id


alfbr:
Thanks, greatly appreciated! Merging.

PR #442, opened 2020-11-24
Test case for flow_onephase_energy

goncalvesmachadoc:
Both simulators work fine now, flow_onephase and flow_onephase_energy.

### DIFF
--- a/spe1/SPE1CASE2_THERMAL_ONEPHASE.DATA
+++ b/spe1/SPE1CASE2_THERMAL_ONEPHASE.DATA
@@ -10,8 +10,6 @@
 -- to simulate thermal water injection/production
 -- in a water single-phase system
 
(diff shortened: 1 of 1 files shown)


atgeirr:
The same questions posed for #441 apply here as well.

atgeirr:
Can the dummy tables here be removed? Can the schedule be shortened?

goncalvesmachadoc:
flow_onephase_energy still has issues, so this case is not ready to go into regression tests.

bska:
@goncalvesmachadoc: Is this the same as the one you added in #500?

goncalvesmachadoc:
Yes, it is. Now the case runs ok.

bska:
> Yes, it is. Now the case runs ok.

Thanks for confirming.  Then I'll just close this as the case is already in the master branch.

PR #445, opened 2020-11-26
Added comments that were erronously removed.

alfbr:
Putting back again comments that dropped out on the latest pull request.


alfbr:
Self-merging as trivial fix. None of the cases are under testing yet.

PR #456, opened 2020-12-04
Add flow dynamic solvent test case

totto82:


alfbr:
Thanks for the contribution, this looks good! In order to merge this we need copyright and license headers also on the include files. Preferable I would also do the following ,but I leave it to you to decide:
-extend the folder name to something understandable, or move the case to the spe5 folder rather than creating a new folder
-add one more line in the DATA-file header to say something descriptive about what dynamic properties are tested (having to look up a publication to know what failed when a test fails is cumbersome)

totto82:
Thanks for the comments. I have followed your suggestions.

bska:
Just one comment: Locally calculated solutions are almost always misleading.  The CI system tends to compute solutions that differ from those calculated on your local machine so we typically end up computing updated reference values the first time we activate a new regression test.

totto82:
> Just one comment: Locally calculated solutions are almost always misleading. The CI system tends to compute solutions that differ from those calculated on your local machine so we typically end up computing updated reference values the first time we activate a new regression test.

Do you want me to remove the reference from this PR? Or can it stay. It will need an update. At least locally it does not go through after I rebased on master.

bska:
>> Locally calculated solutions are almost always misleading.
>
> Do you want me to remove the reference from this PR?

Yes, I think I'd prefer if the solution/output files were **not** part of this PR.

> [The solution] will need an update.

That's an even better reason not to include the output in my opinion.

totto82:
> Yes, I think I'd prefer if the solution/output files were not part of this PR.

Done

bska:
jenkins build this please

alfbr:
Excellent, thank you! Merge when green.

PR #457, opened 2020-12-08
Automatic Reference Data Update for PR https://github.com/OPM/opm-simulators/pull/2922

jenkins4opm:
Reason: https://github.com/OPM/opm-simulators/pull/2922

opm-common     = 1ef1646fb191b71a62dbef184942de9d93a21c8c
opm-grid       = 5343d4635c63a40ca406978f0aecaac5f8815a53
opm-material   = 98c5e38876a436a84c78560230795fd94b1953ab
opm-models     = b980e548e36b3f8a77e1374743ccf4e17b5b532c
opm-simulators = 67ac92dfefb271581d62d99550b188ef57e8070b

### Changed Tests ###

  * spe1
  * spe1_brine
  * spe1_metric_vfp1
  * spe9
  * spe9group
  * udq_uadd
  * udq_undefined
  * udq_in_actionx
  * wsegsicd
  * compl_smry


bska:
Please note that there's a race condition in merging the new reference data at the same time as the upstream changes.  If the integration build of the  upstream changes happens to start before the reference solutions have been successfully installed on the CI system, then the integration build will fail.  Therefore it is **usually** best to merge the reference solutions first and then wait until the CI system confirms that the new solutions have been installed.

The downside to doing it this way is that any PR build checks that start in the window between merging the reference solution and the upstream changes may fail spuriously due to inconsistent reference solutions.

atgeirr:
> Therefore it is **usually** best to merge the reference solutions first and then wait until the CI system confirms that the new solutions have been installed.

Good point. What kind of confirmation were you thinking of?

bska:
> > Therefore it is **usually** best to merge the reference solutions first and then wait until the CI system confirms that the new solutions have been installed.
> 
> Good point. What kind of confirmation were you thinking of?

Looking at the dashboard and manually ensuring that [the opm-tests job](https://ci.opm-project.org/job/opm-tests/) has run.  In this case we happened to win the race because there was a PR build check that happened to run on "Post Builder" which in turn blocked integration builds. We can of course not count on this happening every time.

atgeirr:
> We can of course not count on this happening every time.

Still, that is a useful rule of thumb: If post-builder is busy, merge away!

PR #489, opened 2021-02-26
Add testcase for parallel EQUALS PORV behavior

joakim-hove:


alfbr:
Thanks for the test case! If you want this merged, please add one line in the header explaining that this case tests EQUALS and pore volume (re-phrase to fit if needed). Then feel free to merge.

joakim-hove:
> If you want this merged [...]

Yes I do - thank you! I have added a short description of what the test test is for - merging.

PR #495, opened 2021-03-05
Fix glift

alfbr:
Testing group gas lift case uncovered errors in the input deck. These are hopefully resolved here.


bska:
jenkins build this please

tskille:
Looking good. Non of the models in folder model5 is under regression. Merging now.

PR #497, opened 2021-03-08
New test case for model5

tskille:
- GCONINJE item 10 = NETV
 - 
renaming existing cases since these should build on this case.


alfbr:
Looks good, thanks for the case! Merging.

PR #502, opened 2021-03-13
Automatic Reference Data Update for PR https://github.com/OPM/opm-simulators/pull/3102

jenkins4opm:
Reason: https://github.com/OPM/opm-simulators/pull/3102

opm-common     = 8d11fc5458bc42803354de7d5bc947747c6b6084
opm-grid       = f2cf61cf0bfafa76c11d3b612064bec433d4b2b2
opm-material   = a1dd037984b4f7ee2884d9fa3d21219da13161f1
opm-models     = 6d9c71d54e205705f190ce29ef6102f9ba3f30dd
opm-simulators = 36ad803883a06470888310d33947c36659b52544

### Changed Tests ###

  * spe1_metric_vfp1
  * base_model2
  * multregt_model2
  * multxyz_model2
  * multflt_model2
  * multpvv_model2
  * swatinit_model2
  * endscale_model2
  * hysteresis_model2
  * multiply_tranxyz_model2
  * editnnc_model2
  * model4_group
  * model4_udq_group
  * norne_reperf


bska:
Just for the record, merging new reference solutions at the same time as the PRs that generated those new solutions is almost guaranteed to lead to a race condition.  In this specific instance the upstream PR happened to get into the build queue before this PR which means that the integration build for the upstream PR will fail.

joakim-hove:
> Just for the record, merging new reference solutions at the same time as the PRs that generated those new solutions is almost guaranteed to lead to a race condition

Ehhh - I thought that was the way it should be dine🤔

bska:
> > Just for the record, merging new reference solutions at the same time as the PRs that generated those new solutions is almost guaranteed to lead to a race condition
> 
> Ehhh - I thought that was the way it should be done

Ideally, yes, but that's sadly not the way our system behaves.  The only safe way to merge these updates if we want the integration build to succeed is to merge the reference solution first and then wait until the post-commit build in opm-tests has actually run before merging the PRs that depend on updated reference solutions.  This leaves a time window where the overall system is in an invalid state and will break PR builds, but in my opinion that's better than having failing integration builds.

PR #508, opened 2021-03-23
Add more co2 tests

totto82:
Test case for diffusion in co2-brine system
Test case for dissolution of co2 in brine


totto82:
I want to add them to the Jenkins list before release. They take less than a second to run.

alfbr:
> I want to add them to the Jenkins list before release. They take less than a second to run.

Sure, the more the better. Merging.

PR #519, opened 2021-04-13
Add additional summary variables to ACTIONX_M1 testcase

joakim-hove:
Add new summary variables


joakim-hove:
This is not yet under regression testing - will merge right away; and then finally https://github.com/OPM/opm-simulators/pull/3062 can be merged.

PR #532, opened 2021-05-17
Fix include paths for 4_NUMAQ_MODEL3.DATA.

crayxt:
It seems like include paths for this file should be corrected the same way as it is in other DATA files in the same folder.

Thanks,


bska:
You're right, this is necessary.  I made a similar change to my local copy of the test data, but I was not able to run the model in Flow.  Still, this is good so I'll merge into master.

crayxt:
@bska I haven't run this too, found by tokenizing the data deck.

PR #533, opened 2021-05-18
Add RESV test case

totto82:
A test case that tests
1) RESV in prediction mode for producers
2) RESV for production groups


alfbr:
Thanks for the test-case! Please include a sentence in the file header to inform what the case tests (i.e., RESV group control), and this is ready to go in.

alfbr:
Thank you! Please merge if you are ready at your end.

PR #552, opened 2021-06-11
Add trailing "-- $REPORT_STEP" to DATES keywords

joakim-hove:


joakim-hove:
jenkins build this please

bska (approved):
Thanks.  This looks good to me.

bska:
@akva2 : This PR is ready for merging in my opinion so we can use it as a basis for experimenting with push notifications in opm-tests if you'd like.

akva2:
okay, i don't have access to it so those with access need to configure the webhook on the opm-tests repo, like described here https://stackoverflow.com/questions/48093039/jenkins-github-hook-trigger-for-gitscm-polling @alfbr i guess ?

PR #574, opened 2021-08-03
twophase gas-water test 

plgbrts:
Test case for the  two-phase gas-water functionality. Test case is based on SPE1 CASE2.


alfbr:
Thank you! Merging.

PR #600, opened 2021-09-14
ACTIONX GCONINJE Test

OPMUSER:
Test is based on model2 modified to be a prediction case, with the grid etc. files loaded via the PATH keyword to reduce redundancy.

The model tests the use of the ACTIONX and GCONINJE keywords to control gas and water injection rates.
( 1) Production wells with preferred phase (OIL, GAS)
     OP01  - OIL
     OP02  - OIL
     OP03  - OIL (Completion 2)
( 2) Injection wells with preferred phase (GAS, WATER)
     GI01 - GAS
     WI01 - WAT
( 3) ACTIONX GCONINJE - Re-inject 0.95 of produced gas up 200e3 m3/day
( 4) ACTIONX GCONINJE - Voidage replacement with makeup water with maximum 4000 m3/day.


joakim-hove:
Good. The PR should not include the simulation files - the reference files will be generated by the build/test system from a (already existing) script in opm-simulators.

joakim-hove:
I guess something could be said for having the models self contained - at least within the same directory; i.e. I would be slightly hesitant to use `PATHS`to point *out* of the current directory?

OPMUSER:
I think using PATH is a good idea, as it means we are not duplicating models for different tests. The idea being we can eventually just have a base model directory with all the test feeding off the base model.

joakim-hove:
The `UNSMRY` file is still around ....

PR #603, opened 2021-09-15
Create ACTIONX_GCONPROD.DATA

OPMUSER:
Depletion case with gas field operation.

ACTIONX GCONPROD - Set max gas rate for when GOR >= 150 m3/m3.
ACTIONX GCONPROD - Once GOR >= 200 m3/m3 cut back gas rate by UDA variable.
ACTIONX GCONPROD - Once FOPR <= 500 then convert to gas field.
ACTIONX WCONPROD - Gas field minimum gas rate of 50E3 m3/day then shut-in all wells and exit.


alfbr:
Thank you, this look good! I like that you have spread the different actions in the schedule, and included the full run in one file. This makes the test more comprehensive, and very easy to get a full overview of. Is there any reason why you deviate from copyright notice format of the other decks? I believe it makes sense to keep the same format, or if there are good reasons to change it, we should consider updating all. With an updated header, or some reasoning to change the header format, this goes in.

OPMUSER:
Thank you.

No reason for the different copyright notice in the deck. I have it a template so I just automatically insert it. If you point me to the one you would like to use then I'll update the files.

alfbr:
If you look here, there is an example:
https://github.com/OPM/opm-tests/blob/master/aquifer-num/3D_1AQU_3CELLS.DATA

More precisely, this header:

-- This reservoir simulation deck is made available under the Open Database
-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
-- individual contents of the database are licensed under the Database Contents
-- License: http://opendatacommons.org/licenses/dbcl/1.0/

-- Copyright (C) 2021 Equinor

alfbr:
Thank you! This can now be merged. Just post a note when you deem it ready from your end.

akva2:
it contains merge commits.

alfbr:
> it contains merge commits.

Ah, you are right. The changes to reference solutions should of course be removed.

PR #610, opened 2021-09-17
Add test cases for CO2STORE and aquifers

totto82:
Note. None of this is supported at the moment, but I am working on it.


bska:
For what it's worth I can run `CO2STORE-AQUCT` with current master.  Below is an example plot (gas in place at region level).

---

RGIP-CO2STORE-AQUCT

bska:
@totto82 : Do you want to merge this now?

bska:
> Yes that [merging] would be great. And we should include it as a test I think.

Cool.  I'll merge this now and create regression tests for both `CO2STORE_AQU` and `CO2STORE_AQUCT` as a follow-up.

PR #615, opened 2021-09-24
ACTIONX Tests and Documentation

OPMUSER:
See the ACTIONX.md file for brief documentation and status.


alfbr:
Thanks for the documentation, this is very helpful! Is this ready for review/merge, or are there more commits coming?

OPMUSER:
@alfbr as you can see from the documentation there are few WIP because of bugs, so I don't know what you would like to do in this case. 

For Phase 1 testing I still need to run WEFAC (which found bug #2708) and fixed with OPM/opm-simulators#3546, and make tests for WTEST and UDQ keywords..

alfbr:
> @alfbr as you can see from the documentation there are few WIP because of bugs, so I don't know what you would like to do in this case.

The test cases are of value by themselves. If you are happy with the data-sets as they are now, I suggest we merge. Putting them under regression testing is for another PR in opm-simulators, and that can follow its own time-line.

OPMUSER:
@alfbr yes that is fine they can be merged.

alfbr:
Excellent, let's be on the safe side and give Jenkins a spin first.

alfbr:
jenkins build this please

PR #624, opened 2021-10-04
Output problem of GRUPTREE case has been fixed.

joakim-hove:


joakim-hove:
jenkins bulid this please

bska:
Does "output problem" here refer to OPM/opm-common#2674 to does it refer to something else?

joakim-hove:
> Does "output problem" here refer to OPM/opm-common#2674 or does it refer to something else?

It refers to: https://github.com/OPM/opm-common/pull/2725 - which does not immediately look like an output problem.

joakim-hove:
jenkins build this please

PR #571, opened 2021-07-09
Add summary keyword WMCTL to get well control mode

joakim-hove:
To aid in debugging

PR #635, opened 2021-10-20
Use RPTRST BASIC=2 for entire simulation

joakim-hove:
```
RPTRST
     BASIC=5 /
```
does not work flawlessly ... here I suggest changing to BASIC=2 to simplify restart testing.

@OPMUSER


bska:

As I mentioned in the context of [#571](https://github.com/OPM/opm-tests/pull/571#issuecomment-876974472), we can't update the reference solutions from a PR in opm-tests.

PR #655, opened 2021-11-04
fixed: norne well output

akva2:
newer latex has changed the default output filename to texput.dvi,
not article.dvi.


akva2:
jenkins run norne please

akva2:
well curves are back -> shameless self-merge.

PR #669, opened 2021-11-12
Automatic Reference Data Update for PR https://github.com/OPM/opm-simulators/pull/3695

jenkins4opm:
Reason: PR https://github.com/OPM/opm-simulators/pull/3695

opm-common     = 10314ce169f89e0640f3d74245151817e57674a3
opm-grid       = 54f6fc4defa6710b92eaa07b3311a0dec61162e4
opm-material   = 0267732403438445a7163ac0f96c807e9b45ea3d
opm-models     = 9c5cbc53a302cd104777548dc4531f062171a7f7
opm-simulators = a657cfb5ccdf713eda9f738f92e1335921f4f51b

### Changed Tests ###

  * numerical_aquifer_3d_2aqu
  * numerical_aquifer_3d_1aqu
  * norne_init


GitPaean:
not sure why norne_init here and also not sure why `3D_2AQU_NUM.INIT` is updated.

bska:
> not sure why `3D_2AQU_NUM.INIT` is updated.

Looking at the file difference it appears that the order of `EQLNUM` and `FIPNUM` have been swapped in the new `.INIT` file.  We output those arrays in the order returned by [`FieldProps::keys<int>()`](https://github.com/OPM/opm-common/blob/10314ce169f89e0640f3d74245151817e57674a3/src/opm/output/eclipse/WriteInit.cpp#L335-L336) which is ultimately governed by the [key order](https://github.com/OPM/opm-common/blob/10314ce169f89e0640f3d74245151817e57674a3/src/opm/parser/eclipse/EclipseState/Grid/FieldProps.cpp#L715-L723) of an [`unordered_map<string, FieldData<int>>`](https://github.com/OPM/opm-common/blob/10314ce169f89e0640f3d74245151817e57674a3/src/opm/parser/eclipse/EclipseState/Grid/FieldProps.hpp#L557).  That's not going to be deterministic over time, so I suppose we'll need to impose some kind of order to ensure determinism.

bska:
Merging this to prepare for merging the upstream PR.

PR #672, opened 2021-11-15
Add volve dataset from https://discovervolve.com/

joakim-hove:
This PR adds the VOLVE dataset from https://discovervolve.com/


alfbr:
Copyright and license information seems to be missing completely?

joakim-hove:
> Copyright and license information seems to be missing completely?

It is a section with license and copyright information at the very top of the `VOLVE.DATA` file - but that can certainly be changed.

alfbr:
> > Copyright and license information seems to be missing completely?
> 
> It is a section with license and copyright information at the very top of the `VOLVE.DATA` file - but that can certainly be changed.

Ah, thank you! I should have seen that. It seems like the license notice is wrong or at least strongly misleading. I went to the web-site, and it looks like the license explicitly forbids selling. AFAIK, this means the data set is proprietary and not open. For instance, the open source initiative says this about it (FSF is also crisp clear about it):
https://opensource.org/faq#selling

joakim-hove:
> It seems like the license notice is wrong or at least strongly misleading

I have tried to extract license / copyright information from this page: https://discovervolve.com/citation-non-commerciality-clause/ Do you mean that I have interpreted that web-page wrong, or that I have looked in the wrong place for license information?

> I went to the web-site, and it looks like the license explicitly forbids selling

Maybe we are looking at different sources? Let us be certain that we are looking at the same license information before discussing further.

alfbr:
> I have tried to extract license / copyright information from this page: https://discovervolve.com/citation-non-commerciality-clause/ Do you mean that I have interpreted that web-page wrong, or that I have looked in the wrong place for license information?

We are looking at the same license. You have not included the full license. The problematic part (which is omitted) reads:

> The data shall not be resold.

joakim-hove:
> We are looking at the same license. You have not included the full license. The problematic part (which is omitted) reads:

OK - I understand; not including the full license was an oversight on my part - but I interpret your main point to be: "With the clause which forbids reselling the datasets as a whole is not free and thereby should also not be included in opm-tests?"

alfbr:
> > We are looking at the same license. You have not included the full license. The problematic part (which is omitted) reads:
> 
> OK - I understand; not including the full license was an oversight on my part - but I interpret your main point to be: "With the clause which forbids reselling the datasets as a whole is not free and thereby should also not be included in opm-tests?"

From what I can gather, yes, the data set is neither free nor open. If that is the case, then yes, I believe we should not open that door in opm-tests. Maybe my co-maintainers can chime in with their opinions here @tskille and @bska ?

tskille:
In my opinion we should only include open data sets in opm-tests.

atgeirr:
I *strongly* disagree that the no-sales clause makes this a proprietary dataset, since the license allows free use, modification, sharing and sharing of modifications. These are the same essential freedoms granted by open source licenses for software. I think we should include this!

bska:
Since none of us are copyright law experts I think it's prudent to **not** include this dataset in opm-tests.  The fact that we only have opinion at this point, different opinions at that, suggests that the issue is not clear cut.  I'm not prepared to incur legal costs because we failed to honour a requirement of the license.

At best I think we could add a script, or a description, of how to download the dataset manually and making it runnable.

alfbr:
> I _strongly_ disagree that the no-sales clause makes this a proprietary dataset, since the license allows free use, modification, sharing and sharing of modifications. These are the same essential freedoms granted by open source licenses for software. I think we should include this!

I am simply trying to understand what the authorities say on this topic. From the FSF we have this statement:

> If a license does not permit users to make copies and sell them, it is a nonfree license.

ref. https://www.gnu.org/philosophy/selling.html

and as I already quoted the open source initiative:

> Can I sell Open Source programs? Even if I haven't written it?
>
>    Yes, you can. 

ref. https://opensource.org/faq#selling

Hence, I am a bit in the dark as to what you base your opinion on?

alfbr:
> At best I think we could add a script, or a description, of how to download the dataset manually and making it runnable.

I am not sure a script as viable. From what I can see you need to register with e-mail, full name, affiliation and log-in credentials before you can gain access to the data.

alfbr:
> You can optionally register with email++ if you want to get updates from the discovervolve project, but you are by means required to so in order to download the data.

Not sure I understand you here, maybe you have a misprint somewhere? In any case, do you still want to see this data set included in opm-tests?

atgeirr:
> Since none of us are copyright law experts I think it's prudent to **not** include this dataset in opm-tests. The fact that we only have opinion at this point, different opinions at that, suggests that the issue is not clear cut. I'm not prepared to incur legal costs because we failed to honour a requirement of the license.
> 
> At best I think we could add a script, or a description, of how to download the dataset manually and making it runnable.

I think this is a misunderstanding. We are clearly *allowed to* share it, there is no reason to share a script instead. I do not think @alfbr for example intends to say something else?

> I am simply trying to understand what the authorities say on this topic. From the FSF we have this statement: [...]

The question is if we *want to* share it. Some say no, as this is proprietary, not free or open etc. These opinions are based on what organizations like the FSF say about the topic. As I wrote above, the license does allow for the four central freedoms. I will admit that according to the definitions given there, it can be argued that the Volve data set is still not Free, due to the "no sales" clause. (Although calling it proprietary is not warranted, I think.) My argument is that this clause poses little or no restrictions for users of or contributors to OPM. As far as I can tell, no-one have wanted to "sell" any other data sets, even though we are allowed to. The only restriction that this poses that I think may have merit, is that we would not be allowed to charge a distribution fee for covering storage/computer costs associated with the data.

In summary, I believe there are benefits for the entire OPM community to including this, and no notable drawbacks.

joakim-hove:
> Not sure I understand you here, maybe you have a misprint somewhere?

What I am saying is just that you can download the Volve reservoir model from: https://discovervolve.com/2021/09/25/how-to-run-the-volve-oilfield-dynamic-model-in-opm-flow/ without registering in any way. 

>  In any case, do you still want to see this data set included in opm-tests?

Well - my initial thinking was that more cases which could be run in automatic testing was a good thing for opm/flow, so on that basis I wanted to include the VOLVE dataset into opm-tests. As I see it there are two questions:

1. Is it legally OK to distribute the VOLVE dataset through opm-tests?
2. Do we *want* to distribute the VOLVE dataset through opm-tests?

My position on these is as follows:

1. Yes - we are allowed to distribute the VOLVE dataset through opm-tests. The fact that we, or those downloading from opm-tests, are not allowed to *resell* the dataset does not affect our right to make the dataset available in the form suggested on opm-tests. I acknowledge that the no resell clause makes the dataset not Free&trade; - but that in itself does not limit our possibility to distribute it (of course with the no resell clause intact).
2. Yes - I think we should include it, others disagree.

I don't really have anything more to add to this discussion, I will step back and see what happens.

bska:
> The question is if we _want to_ share it. Some say no, as this is proprietary, not free or open etc. These opinions are based on what organizations like the FSF say about the topic.

For what it's worth, those are the primary authorities on this particular topic.  I'll also note that if we do include anything that's not completely free into our repositories then we also exclude those users/contributors who are in organisations that do not permit anything non-free into their systems.  I have no idea if we have any of those or if we might have any in the future, but I do think it's worth considering.

alfbr:
> For what it's worth, those are the primary authorities on this particular topic. I'll also note that if we do include anything that's not completely free into our repositories then we also exclude those users/contributors who are in organisations that do not permit anything non-free into their systems. I have no idea if we have any of those or if we might have any in the future, but I do think it's worth considering.

Thank you! It actually saddens me that we even have to argue that point.

atgeirr:
> For what it's worth, those are the primary authorities on this particular topic. 

I think this is a valid point, and I have come to the conclusion that it must be guiding here. On our website we write "OPM coordinates collaborative software development, maintains and distributes open-source software and open data sets, and seeks to ensure that these are available under a free license in a long-term perspective." So it follows that our data repositories should be 100% free, and not just 95%. It is reasonable will be expected by us that we follow the definitions of open used elsewhere in the open source and free software community, and the data set in question is not, by those definitions.

Although I think it would have value, I will therefore close this PR.

PR #683, opened 2021-11-22
Nopinch

jilocode:
Added license lines


alfbr:
For some reason you have added the  model5/4_GLIFT_MODEL5.DATA file change to this branch now. In any case, that change is already merged, and there is according to github no conflicts. I am merging this now, we should be able to resolve any problems if they arise.

PR #689, opened 2021-11-29
Update Norne restart file

joakim-hove:
Update NORNE restart file: see https://github.com/OPM/opm-simulators/pull/3708


alfbr:
Update at your convenience.

PR #692, opened 2021-12-04
test case with passive water tracer

tskille:


alfbr:
Thanks for the contribution, merging.

PR #694, opened 2021-12-06
Add CO2STORE energy test case

totto82:
Add a test case for CO2STORE + THERMAL
The plan is to add it as a test case.


alfbr:
Thank you!

PR #698, opened 2021-12-08
Create flushing test case

goncalvesmachadoc:


alfbr:
This looks good from my end, just give a ping when you want it in.

PR #699, opened 2021-12-08
add gas-water test case

goncalvesmachadoc:


alfbr:
Thanks for the contribution! Merging.

PR #730, opened 2022-02-21
Add BPR and BTCNFHEA

goncalvesmachadoc:
Test for BPR and BTCNFHEA when no oil is present.


bska:
jenkins build this opm-simulators=3820 please

PR #750, opened 2022-04-20
SPE02 Radial and Spider Grid Tests

OPMUSER:
Various radial and spider grid tests to check various input keywords based on SPE comparative solution number two (SPE02) .

Only two radial models run and none of the spider grid models run.


alfbr:
jenkins build this please

alfbr:
Thanks for extending the data sets for radial and spider grids, solid regression testing here is very welcome!

PR #770, opened 2022-06-14
shut two segments using WSEGVALV

totto82:


alfbr:
jenkins build this please

totto82:
Yes, it is ready from my side.  This test case is not currently part of the test suite. I think it should be. I will open a PR to opm-simulators and discuss it with @tskille and @GitPaean

alfbr:
Sounds good, in it goes!

PR #775, opened 2022-06-29
remove RUNSUM from BO_DIFFUSE_CASE1

totto82:
as suggested in https://github.com/OPM/opm-simulators/pull/3956#issuecomment-1167062477


totto82:
This is trivial. Will self merge

PR #780, opened 2022-07-28
Update Runs and Documentation

OPMUSER:
Update runs and documentation, together with additional commentary.


blattms:
Wasn't aware that we binary odp files in here.

OPMUSER:
Yes we had a discussion on this for the GASLIFT tests and on whether to use PDF or ODP file types. We elected to go with ODP as one can edit these and make notes. ODP files are just zip files.

I think its good to see the comparison plots at the time of testing, especially when we are not matching with E100.

PR #783, opened 2022-08-05
SPE10 Tests Update

OPMUSER:
Update SPE10 cases for model 1 tests. Fixed typos in original cases and created two-phase cases. All four cases now give virtually the same results, and the cumulative oil is now similar to the published cases of around 45 Mstb.

Wasted time on trying to get the commercial simulator to run the first two cases to no available, as the producer gets shut-in at around 2,000 days.

I have placed the cases in a separate directory, and will also add the model two cases in this directory as well when they are ready.

See SPE10.md for case description and commentary.


bska:
Just out of curiosity: Why do we need the `SPE10-MOD01-GRID.inc` include file?  The 100-by-1-by-20 cell model 1 is **really** simple so I don't immediately understand why it's modelled using a corner-point description.

OPMUSER:
> Just out of curiosity: Why do we need the SPE10-MOD01-GRID.inc include file? The 100-by-1-by-20 cell model 1 is really simple so I don't immediately understand why it's modelled using a corner-point description.

In the original directory we had both Cartesian regular grid and Corner-Point regular grid models, I just duplicated those tests.

I think this may be because of the OLDTRAN/NEWTRAN difference as the original cases did not match the results in the original paper by quite a bit. The reason for the miss-match was due to typos in the original cases. In particular, the gas injection rate. There was also a typo in the wellbore diameter as well, but this does not make much difference.

alfbr:
Great to see the SPE10 case maturing! Do you want this reviewed/merged now, or do you want to continue working on he Model 2 case?

OPMUSER:
Hold for now. 

Model two now runs as well thanks to some guidance from @atgeirr - takes a while on my machine to run but it at least it now runs to completion. I'm waiting on @atgeirr to provide the results from the paper so we can compare with the published results. A quick review looks like the model is within the range of the published results. 


image

image

I think we should use one of each test case model as part of the release testing, not the Jenkins regression testing, as they take some time run.
 
One question, do we stop at the fine grid model or do you wish to proceed to the up-scaling as well. I think stopping with the fine grid model is reasonable, especially if we get comparable results with the fine grid model.

alfbr:
> Hold for now.

Will do.

> I think we should use one of each test case model as part of the release testing, not the Jenkins regression testing, as they take some time run.

That sounds like the right thing to do, so let's include them on the release testing models.

> One question, do we stop at the fine grid model or do you wish to proceed to the up-scaling as well. I think stopping with the fine grid model is reasonable, especially if we get comparable results with the fine grid model.

I need to refresh my knowledge of these models to give an informed answer, but we can at least start with the fine grid model.

OPMUSER:
Model two cases and updated documentation, ready for review.

bska:
> Model two cases and updated documentation, ready for review.

The model two data files both refer to a few [include files](https://github.com/OPM/opm-tests/blob/bafb25ef1cd5aed51798656718298704ab501522/spe10/SPE10-MOD02-02.DATA#L197-L208) that appear to be missing here.  Are the include paths here entered as intended?

OPMUSER:
> The model two data files both refer to a few [include files](https://github.com/OPM/opm-tests/blob/bafb25ef1cd5aed51798656718298704ab501522/spe10/SPE10-MOD02-02.DATA#L197-L208) that appear to be missing here. Are the include paths here entered as intended?

@bska sorry about that just now uploaded them.

bska:
> just now uploaded [the missing include files]

Thanks.  That said, might it make sense to reference the include files that already exist in [`opm-tests/spe10model2`](https://github.com/OPM/opm-tests/tree/295eca9a1faed4e0614919ac1718b79140ee007b/spe10model2) instead?  That way we could keep a single copy of these relatively large property files.

OPMUSER:
> Thanks. That said, might it make sense to reference the include files that already exist in [opm-tests/spe10model2](https://github.com/OPM/opm-tests/tree/295eca9a1faed4e0614919ac1718b79140ee007b/spe10model2) instead? That way we could keep a single copy of these relatively large property files.

@bska No we should not do this. First of all the PERM array has been modified from the original in [opm-tests/spe10model2](https://github.com/OPM/opm-tests/tree/295eca9a1faed4e0614919ac1718b79140ee007b/spe10model2) and the new cases use the original. Secondly, the intent is to remove [opm-tests/spe10model2](https://github.com/OPM/opm-tests/tree/295eca9a1faed4e0614919ac1718b79140ee007b/spe10model2) and model one and use the new cases. I just don't know how to do that without upsetting Jenkins

OPMUSER:
Ready for review.

If Jenkins does not use https://github.com/OPM/opm-tests/tree/master/spe10model1 and  [opm-tests/spe10model2](https://github.com/OPM/opm-tests/tree/295eca9a1faed4e0614919ac1718b79140ee007b/spe10model2) for regression testing then these two directories can be deleted and the tests in this PR should be used for release testing.

alfbr:
jenkins build this please

bska:
> If Jenkins does not use spe10model1 and spe10model2 for regression testing

These models are indeed not part of the simulator's regression testing, but they **are** used in a kind of [consistency check](https://github.com/OPM/opm-common/blob/dacd635c381830442d5172ab2027cff59ec6355d/ExtraTests.cmake#L107-L108) for the parser layer.  That said, that consistency check is, by now, of very little value in isolation.  It uses the [`parse_write`](https://github.com/OPM/opm-common/blob/dacd635c381830442d5172ab2027cff59ec6355d/tests/parser/integration/parse_write.cpp) utility which essentially performs the following steps

1. Read an input file and create a `Deck` structure
2. Create the fundamental `EclipseState`, `Schedule` and `SummaryConfig` objects from this `Deck` structure
3. Serialise the `Deck` structure to a string and read that string back in to another `Deck` structure
4. Confirm that the two `Deck` structures are the same

Note especially that the `EclipseState`, `Schedule` and `SummaryConfig` objects are never used after they're created so there's no check that they contain the expected values.

I can use the use of `parse_write` as a simple debugging tool, but I don't think we should use it for unit testing anymore and I'd be in favour of removing all of those "tests".

alfbr:
jenkins build this please

bska:
> merged commit a27fb50

I'll make an emergency fix to restore the build.

alfbr:
What did I miss, Jenkins seemed content?

bska:
> What did I miss

The last commit which removed the existing, modified, copies of the SPE 10 input decks.

> Jenkins seemed content?

Yes.  The last commit was never run through Jenkins.

alfbr:
> > Jenkins seemed content?
> 
> Yes. The last commit was never run through Jenkins.

I see, I did rerun Jenkins after that commit, but I guess it somehow was not updated.

bska:
> > > Jenkins seemed content?
> > 
> > Yes. The last commit was never run through Jenkins.
> 
> I see, I did rerun Jenkins after that commit, but I guess it somehow was not updated.

No, it's the subtle wrinkle in the Jenkins setup that builds triggered from a downstream repository don't run unit or regression tests in upstream repositories.  In this case, the removed input decks were only used in opm-common and those unit tests did not run as part of the last build check.

alfbr:
> > > > Jenkins seemed content?
> > > 
> > > 
> > > Yes. The last commit was never run through Jenkins.
> > 
> > 
> > I see, I did rerun Jenkins after that commit, but I guess it somehow was not updated.
> 
> No, it's the subtle wrinkle in the Jenkins setup that builds triggered from a downstream repository don't run unit or regression tests in upstream repositories. In this case, the removed input decks were only used in opm-common and those unit tests did not run as part of the last build check.

Ah, that makes sense. Thanks for the quick fix!

OPMUSER:
@bska Many thanks for fixing this.

PR #784, opened 2022-08-09
Import Single-Phase, High Rock Compressibility Water Case

bska:
Provided by Edmund Stephens.  Uncovered a problem in Flow's calculation of averaged quantities like the field-level pressure value (`FPR`) in models without hydrocarbons and high compressibility.

See Issue OPM/opm-simulators#3984 for additional details.


bska:
jenkins build this please

alfbr:
Thanks for the contribution! This is good to go from my end, but please check that the change on .gitignore is intentional.

bska:
> please check that the change on .gitignore is intentional.

Good point.  Those changes are indeed intentional and intended to avoid accidentally importing Emacs auto-save files.  I added that change because I did come very close to importing a `*~` file here.

PR #787, opened 2022-08-12
Output Selected Connection Transmissiblity Factors

bska:
In preparation of adding a regression test based on this model.

While here, also switch to Unix EOL and switch of the "executable" bit.  Input data does not need to be marked as executable.

### DIFF
--- a/actionx/ACTIONX_WPIMULT.DATA
+++ b/actionx/ACTIONX_WPIMULT.DATA
@@ -376,6 +376,13 @@ WGITH
 /
 WWPTH
 /
+--
+--      CONNECTION TRANSMISSIBILITY FACTORS FOR SELECTED CONNECTIONS
+--
+CTFAC
+PROD1 6  3 7 /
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

alfbr:
The change looks fine. However, I am puzzled that the diff is the whole file. Merge at your convenience.

bska:
> I am puzzled that the diff is the whole file.

That's the Windows (CRLF) to Unix (LF) EOL markings.  If you'd like to see just the new lines you can turn on "whitespace filtering" in the diff view, but it boils down to
```diff
+ --
+ --      CONNECTION TRANSMISSIBILITY FACTORS FOR SELECTED CONNECTIONS
+ --
+ CTFAC
+ PROD1 6  3 7 /
+ INJ1  2 13 1 /
+ /
```

> Merge at your convenience.

Thanks.  I'll merge this now.

PR #799, opened 2022-08-29
Test case with API tracking

tskille:
Model set up with PVTO (live oil), PVDG (dead gas) and water. API tracking with 2 APIGROUPs.

This is a model for testing API tracking. The reservoir contains a southern part with light oil (high API and low viscosity) and a northern part with heavy oil (low API and high viscosity). There are 4 producers located in the northern part and 4 injectors located in the southern part, and injection is controlled by voidage replacement. A total of 3 M Sm3 of light oil is flowing from south to north causing API to raise over time in the producers.


alfbr:
Thank you, test on API-tracking is very valuable now that we plan to implement it. All good from my end, merging.

PR #807, opened 2022-09-21
WVFPEXP: Documentation via Markdown Instead of ODT File

OPMUSER:
Here we generate the results in OPM ResInsight via PNG files and then use markdown to organize the results. The intent, if everyone agrees, is to replace the the ODT file with this PR.

I have placed all the plots in the one markdown file, but I could also place each case in a separate file, if that is less complicated to review the results.

Note that there a few limitations in OPM ResInsight at the moment that prevents the generation of the plots being fully automatic. Nevertheless, once the plots are setup generating a new version based on the master being updated is virtually automatic.

Comments appreciated.


alfbr:
Thanks for the contribution! This sounds very interesting, so please allow a few days to get an overview of what this entails.

OPMUSER:
Yes, I thought you might find this interesting.

Basically, now that I have figured out how all this works, I have a Python script that generates all the plots in an OPM ResInsight project file. So once you set up one project file with the desired cases and plots, it is relatively quick and easy to create the project files for the other cases. One can then use the script to export all the plots for a given case. 

As I mentioned there are few limitations with the current Python scripting facility, that I had to work around, but once done, this is very efficient at documenting the results compared to copying and pasting the results into an ODT file.

Further automation may also be possible by creating the project files from a master project file, but I'll HOLD on that for now.

blattms:
Nice! :+1:

OPMUSER:
> That point is to replace the simulation from Eclipse with a known to be correct simulation from Flow. There are many reasons for this. The fundamental one being that Eclipse solutions are not reproducible. Very few have the license, and if they do, they probably do not have the same version as the solution was created with. Second to that, Eclipse and Flow solutions are not directly comparable on the benchmark cases, just the time stepping differences make them often very different. Lastly, at the point when a Flow simulation is known to be correct, that should suffice as the "truth", so I do not see any added value in Eclipse solutions beyond that point.

I don't have a problem with having the E100  "base" run, being replaced by an OPM Flow run, just that we should do the change for all the cases in an opm-test directory at the same time. So that we don't mix and match the different sources. Perhaps we should do this as part or the release cycle, and suffix the cases with -2210, for example, instead of ECL.

> What I do not understand is exactly how this will be updated. Are you planning to do it at your end with some automated set-up, or is there something in the pull request to facilitate easy update? Secondly, I wonder exactly what was not possible to automate from Resinsight? Maybe it is something we should look closer into, after all, it looks like we could even have Jenkins help us update these plots automatically.

Previously, I loaded the cases into OPM ResInsight generated the plots, then copied and pasted them into LibreOffice. This was very time-consuming. What I do now is

1. I set up an OPM ResInsight project file for one case, with the plots generated via plot templates files. Only need to this once.
2. For the other cases, I use the previous created OPM ResInsight project file and replace the summary files with the new cases.
3. Once all the OPM Resinsight project files are created, then we can use the Python script to generate the plots. I could extend the Python script to automatically load the project files and generate the plots, but I was waiting on feedback before expanding the script.
4. When we run new cases, we just have to repeat (3) to get the new plots.

At the moment we cannot fully automate this, due to some limitations in the OPM ResInsight Python interface, so I have to generate the plots locally and then upload them to GitHub. But I see this as a temporary solution for the time being.However, anybody can do the workflow once the OPM ResInsight project is setup.

Some of the Python interface limitations are:

1. One has to create the plots for each well in the project manually, as there is no way to generate the plots for all the wells in the project, ideally with different plots for producers and injectors. So it would be useful to be able to loop over wells for all well plots.
2. Another limitation is that fact that you cannot use an object name in the plot title, %WELL% for well name for example. This is a pain because the plot title is used as the file name, and consequently files will be overwritten if the plot title is the same. There is an auto plot title option, but then you just get the well name, and if you have several plots for a well, then you have the overwriting problem again.
3. Would be useful to be able to load plot templates from the Python interface.

So can we fully automate this with Jenkins? Not now, but that should be the aim, and we are getting there. There is now a way to run OPM ResInsight in headless mode, so that is a big step forward to incorporate automatic plot updates with Jenkins. How I think this would work? We would set up a master OPM ResInsight project file for a test directory, then the system would call OPM ResInsight in headless mode, load a case, generate all the plots by looping over the well plots, group plots etc., export the plots, load next case and repeat. That is what we should aim to have in the near future, if everybody agrees that is the way to go.

alfbr:
> I don't have a problem with having the E100 "base" run, being replaced by an OPM Flow run, just that we should do the change for all the cases in an opm-test directory at the same time.

Sounds good! Then I am happy with that clarification, we just need to work out how we add a reference Flow simulation. We do not have to do that now.

> At the moment we cannot fully automate this, due to some limitations in the OPM ResInsight Python interface, so I have to generate the plots locally and then upload them to GitHub.

As long as you are happy with that, I will not complain :-) Yes, I do think we should look into automating it, but again, this can be for another PR. Hence, I am merging now.

PR #809, opened 2022-09-23
Adds a test case for testing KRNUM

hakonhagland:
This test is created from SPE1CASE2.DATA by adding a second saturation function that has a lower oil relative permeability. This saturation function is used in the X-direction by applying the KRNUMX keyword. See https://github.com/OPM/opm-common/pull/3113 for more information.

### DIFF
--- a/spe1/SPE1CASE2_KRNUM.DATA
+++ b/spe1/SPE1CASE2_KRNUM.DATA
@@ -161,6 +161,8 @@ SGOF
 0.88	1.0	0.000	0 /
 0	    0	  0.5	  0
 0.88	1.0	0.000	0 /
+0	    0	  0.75	  0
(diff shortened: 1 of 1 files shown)


bska:
Does this model run with your implementation of directional relative permeability?  The reason I ask is that there's no `TABDIMS` keyword here so the parser will assume that there is exactly one tabulated saturation function.

hakonhagland:
> Please included one sentence in the header (right after copyright) where you explain what this case tests

@alfbr See update

> consider expanding to three different tables instead of using the same table for Y- and Z-direction, that enables testing all directions.

@alfbr Good idea! See update.

> that there's no TABDIMS keyword here

@bska The keyword is there, am I missing something?

alfbr:
Excellent, thank you!

PR #811, opened 2022-10-02
test model with api tracing, dead oil

tskille:
Model set up with PVDO (dead oil), PVDG (dead gas) and water. API tracking with 2 APIGROUPs.


bska:
Thanks a lot for this model, really appreciated.  I notice that there are number of `INCLUDE` statements for which the corresponding include files are not part of this PR.  Should those be added with `PATHS` instead?

tskille:
This is the second model located in folder opm-tests/api_tracking. Notice that this model is using the same include files as the model already in the folder except for the pvdo include file. This means that the other include files for this model is already there. Hence, I don't see the need for using the PATHS keyword.

bska:
> This is the second model located in folder [`opm-tests/api_tracking`](https://github.com/OPM/opm-tests/tree/336549b8e6928762a6bf86ddc16c8300dcfea7c4/api_tracking). Notice that this model is using the same include files as the model already in the folder, except for the pvdo include file.

Ah, right.  Thanks for pointing that out as I didn't notice that the first time around.

bska (approved):
Thank you for the clarification.  I'll merge into master.

PR #812, opened 2022-10-02
changing from \ to / in PATH and INCLUDE

tskille:
data file GASLIFT-03.DATA is not running with commercial simulator without these changes


tskille:
I see that "correct" slash has been used for the rest of the data decks GASLIFT-XX.DATA.

alfbr:
Thanks for the clean-up! Merging.

PR #814, opened 2022-10-03
GASLIFT: Re-run Cases and Update Documentation

OPMUSER:
Re-ran all the gas lift cases and updated the results.

Deleted the ODP file and replaced with markdown and *.png graphics files.

Note the generation of the plots and the markdown files are now automatic,  so re-generating the plots after a master update, now only takes a few minutes. However, reviewing the results and updating the table summary takes some time.


alfbr:
Thanks for the update! This is agreed to be a big step in the right direction, so I am merging this as is.

PR #819, opened 2022-10-06
Group Control Test Cases

OPMUSER:
This set of cases are based on the existing MODEL2 test cases, which are being re-factored for clarity and for automatic generation of the resulting plots. Note that not all cases will necessarily be used and re-factored during the review. The intent is to remove the existing cases from MODEL2 directory after the re-factoring is complete. Cases not re-factored will remain in the MODEL2 directory.

This PR is WIP until all the cases have been completed.

Comments and feedback warmly appreciated.


OPMUSER:
Ready for review and comments.

bska (approved):
Thanks.  I'll merge this into the master branch now.

PR #826, opened 2022-10-18
3 test cases for krnum functionality

tskille:
These models have focus on directional relperm for SWOF (oil water system). 

```
KRNUM-02X.DATA   dominating flow direction is X
KRNUM-02Y.DATA   dominating flow direction is Y
KRNUM-02Z.DATA   dominating flow direction is Z
```
Descriptions of the models can be found in the header of the DATA files


alfbr:
Thank you, this extends the testing done in the previous test model. Merging.

PR #834, opened 2022-10-26
Add case testing MULTFLT during SCHEDULE.

atgeirr:
Same as the FAULT_MODEL_1.DATA case, but changing the fault transmissibility multiplier during the run.

### DIFF
--- a/model1/MULTFLT_MODEL_1.DATA
+++ b/model1/MULTFLT_MODEL_1.DATA
@@ -1,0 +1,377 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2018, 2022 Equinor
+
+-- Testing the FAULTS keyword and MULTFLT keyword.
+-- Compared to the FAULTS_MODEL_1.DATA case, this case
+-- also tests MULTFLT during the SCHEDULE.
+
+-- This case has multisegment wells. 
+ 
+-- 3 wells in the beginning, 3 more are turned on. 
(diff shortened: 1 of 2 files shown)


atgeirr:
I was not aware of the other test. And you are almost right: it is under testing, but not for regressions. It is only tested that it produces equivalent results in parallel and serial runs. It is not actually verified against a known good reference solution.

Furthermore, the changes in MULTFLT (in 3_A_....) come somewhat late in the schedule, and are very small (from 1 to 0.1-0.4), which does not perturb the solution a lot. The current proposed test here makes a much more dramatic change (applying a multplier of 100.0) which causes a much greater difference between the solutions.

I will close this PR, and instead add a regression test for the 3_A_MPI_MULTFLT_SCHED_MODEL2.DATA case. However, I will also make that case more extreme by changing the MULTFLT keyword in its schedule so that we will be able to distinguish behaviour.

A note on background: release testing of a field case using this feature revealed that the new fast assembly did not handle this correctly. The test is intended to prevent future failure to honor the feature. (Recall that the existing tests did not.)

alfbr:
> I was not aware of the other test. And you are almost right: it is under testing, but not for regressions. It is only tested that it produces equivalent results in parallel and serial runs. It is not actually verified against a known good reference solution.

Nice catch! You are right, and yes, it really should be under regression testing. This an oversight when the test was added.

> Furthermore, the changes in MULTFLT (in 3_A_....) come somewhat late in the schedule, and are very small (from 1 to 0.1-0.4), which does not perturb the solution a lot. The current proposed test here makes a much more dramatic change (applying a multplier of 100.0) which causes a much greater difference between the solutions.

This is a welcome change. Large impact is always a good thing for regression tests.
>
> I will close this PR, and instead add a regression test for the 3_A_MPI_MULTFLT_SCHED_MODEL2.DATA case. However, I will also make that case more extreme by changing the MULTFLT keyword in its schedule so that we will be able to distinguish behaviour.

Thanks, this i highly appreciated.

PR #835, opened 2022-10-26
Added the Drogon reference data set

alfbr:
This adds the Drogon reference data set to opm-tests. It is intended for the integration testing done with benchmarking, not the continuous regression testing.


bska (commented):
I notice that the following include files are unused.  Is that intentional?  I also wondered why there were a number of unformatted files here, but then I noticed that they are used in `IMPORT` rather than `INCLUDE` so that's fine.

* [schedule/55_33-1.wells](https://github.com/OPM/opm-tests/blob/8f0f4b2ee5308415ef04c5a31d7520acfb2e8d06/drogon/include/schedule/55_33-1.wells)
* [schedule/55_33-1.sch](https://github.com/OPM/opm-tests/blob/8f0f4b2ee5308415ef04c5a31d7520acfb2e8d06/drogon/include/schedule/55_33-1.sch)

alfbr:
Thanks for the thourough review! I have addressed all comments, please give it another go.

bska:
> have addressed all comments, please give it another go.

Perfect!  Thanks a lot for the updates.  Below's a plot of the average field pressure from a local run of the `_HIST` model with 14 MPI processes.

---

DROGON_HIST_Pressure_average_value

bska:
jenkins build this please

bska:
Just a word of caution: The `_PRED` case does not run.  While there may be more issues, I could not get past an unexpected/templatised [`RESTART`](https://github.com/OPM/opm-tests/blob/8a08456e5695e631778d9e6fb075487183bfa1cf/drogon/model/DROGON_PRED.DATA#L198-L199) specification.  Is that intentional?  If not, do we want to address that before merging or is that better suited to follow-up work?

alfbr:
> Just a word of caution: The `_PRED` case does not run.

Thanks for checking, that is indeed unintentional. The original PRED file snuck back in, fixed now.

bska:
> > Just a word of caution: The `_PRED` case does not run.
> 
> Thanks for checking, that is indeed unintentional. The original PRED file snuck back in, fixed now.

Confirmed, thanks for fixing.  Both the `_HIST` and the `_PRED` cases run now in local testing (`FPR` below).  If you would like to clean up the commit history here, e.g., to squash commits and/or rebase onto the current master sources then now is good time.  Otherwise, I think we should merge the current PR.

---

Pressure_average_value

alfbr:
All squashed to one.

bska:
> All squashed to one.

Perfect!  I'll merge this into master.  Thanks a lot for expanding our set of field cases, this is greatly appreciated!

PR #837, opened 2022-10-26
adapt test cast to new order of items

totto82:
Changes to BC suggested by @OPMUSER


bska (approved):
Merging this now in preparation of merging the upstream PR.  Thanks a lot for the update.

PR #840, opened 2022-10-28
Gaslift: Documentation Update

OPMUSER:
I have re-factored the documentation for the gas lift cases. I have not changed the DATA files or the resulting plots. This is just to make it easier to review the results for a particular case.


OPMUSER:
Ready for review.

alfbr:
It seems the changes are too big for github, I am not able to check the pull request through the web-interface. I assume this is due to the large number of binary files. The pull request itself seems to touch 268 files. It is possible to download the pull request locally, and go through the changes, but it is of course less efficient. If this is a recurring issue, we may want to rethink the set-up.

OPMUSER:
> It seems the changes are too big for github, I am not able to check the pull request through the web-interface. I assume this is due to the large number of binary files. The pull request itself seems to touch 268 files. It is possible to download the pull request locally, and go through the changes, but it is of course less efficient. If this is a recurring issue, we may want to rethink the set-up.

@alfbr Yes, this because there are a total of 15 gas lift cases, with about 16 to 19 plots per case, depending on the scenario.  So it is the number of cases, which is abnormally high, combined with the various plots. We could eliminated some of the plots if required, for example the Pressure_Comparison plots for the wells. Open to suggestions.

I use the Git command line or the GitHub Desktop application to sync between local and remote repositories, so I don't have any problems uploading or downloading.

OPMUSER:
Closing as will be replaced with GASLIFT2 with fewer plots.

PR #841, opened 2022-10-31
SPE10 Updated Documentation

OPMUSER:
Remove ODP file documentation and replace with Markdown documentation, contents are virtually the same.

Ready for review.


spe10/SPE10.md
```diff
@@ -1,19 +1,88 @@
...
+   SPE, August 2001.
+
+Originally, these tests were derived when OPM Flow was only able to run three-phase models, and thus these tests were
+formulated as pseudo two-phase cases. For 2022, the tests have been re-factored with corrections, and OPM Flow results
+have been compared with the published results. Both two-phase and three-phase OPM Flow cases give exactly the same
+results, and the results are comparable to the published results from various vendors.
+
+**Notes:**
+
+1. _Test Type_ column shows if the the case is used for intergration testing (_Int_), or regression teststing (_Reg_)
```
bska, line 25: Tiny nit: "... if THE THE case is ..."

bska (commented):
It's probably not a change from earlier versions of this document, but what are the conditions on republishing simulation results of other vendors/simulators?

spe10/SPE10.md
OPMUSER: @bska I don't know for sure, so I'll defer to someone wiser than me. Note that none of the plots are taken from the actual papers, the data was digitized and the plots in the documentation were derived from the digitized data. Since we state the source, then I don't think it is an issue. But like I said, I'll defer to someone wiser.

Will fix the typo.

OPMUSER:
This has been sitting for a while, can it go in, or are there any change requests?

bska:
> This has been sitting for a while, can it go in,

The PR is blocked by deciding whether or not it is okay to republish simulation results taken from the SPE paper.

OPMUSER:
> The PR is blocked by deciding whether or not it is okay to republish simulation results taken from the SPE paper.

@bska Well we do this all the time, as for example in:
2022-11-29 17_39_18-The Open Porous Media Flow Rese - Atgeirr Flo Rasmussen pdf - Adobe Acrobat

Where we compare SPE1, SPE3, SPE5 ans SPE9. So I don't see an issue with this.

alfbr:
My take on it is that if in doubt, we do not republish copyrighted material. Two wrongs do not make a right. As I see it, it is a learning process where we all do our best, and where we have stumbled in the past. I believe the easiest solution here is simply remove the plots that are copyrighted by others. One may link to them instead. Alternatively, somebody should dig up the reproduction requirements of the copyrighted material. Personally, I have spent way too much time reading through such documents, and hope to avoid it this time around.

spe10/SPE10.md
```diff
@@ -1,19 +1,88 @@
...
+SPE10-MOD01-02| Two-Phase, with Cartesian regular grid.          | MODEL-01   | Int  | Yes     | Results are comparable to other vendors as presented in the paper; however, E100 fails to run to completion.
+SPE10-MOD01-03| Three-Phase, with Corner-Point regular grid.     | MODEL-01   |      | Yes     | Results are comparable to other vendors as presented in the paper. E100 not tested.
+SPE10-MOD01-04| Two-Phase, with Corner-Point regular grid.       | MODEL-01   |      | Yes     | Results are comparable to other vendors as presented in the paper. E100 not tested.
+SPE10-MOD02-01| Three-Phase, with Cartesian regular grid.        | MODEL-02   |      | Yes     | Results are comparable to other vendors as presented in the paper. E100 not tested.
+SPE10-MOD02-02| Two-Phase, with Cartesian regular grid.          | MODEL-02   | Int  | Yes     | Results are comparable to other vendors as presented in the paper. E100 not tested.
+
+Both models are based on the SPE 10 comparative SPE paper:
+
+   "Tenth SPE comparative solution project: a comparison of upscaling techniques"
+   By Mike Christie (Heriot-Wat University) and Martin Blunt (Imperial College)
```
bska, line 15: I guess if we want to be **really** picky it should be Heriot-Watt (two 't's) here.

bska (commented):
Thanks a lot for the updates.  With one minor provision I'm prepared to merge this now.

spe10/SPE10.md
OPMUSER: Done both in Markdown and decks.

bska (approved):
Thanks a lot for the updates.  I'll merge this into master now.

PR #844, opened 2022-11-02
Test case for solvent and directional permeabilities

hakonhagland:
Adds a new test case for solvent and directional permeabilities. This is the same test case as [SPE1CASE2_SOLVENT.DATA](https://github.com/OPM/opm-tests/blob/master/spe1_solvent/SPE1CASE2_SOLVENT.DATA) except that a second SATNUM region has been added by copying the original saturation tables and modifying the saturation table in SOF3 slightly for the second region. The relative permeabilities in the x- and y-directions then use this table instead of the original table.


atgeirr:
Seems like a good idea. Only one question: is the addition of the second SATNUM region sufficient to change the solution significantly, i.e. it will be easy to see if the directional permeability feature is broken?

alfbr:
This is welcome. Like @atgeirr suggests, it is always a good idea to ensure that the tested feature has a significant impact on the solution. However, you may choose to do that if and when you set it up as a regression test. The data set is ready to go in as is. I have one wish though, and that is moving it to the new krnum folder. We hope to refactor the folder structure in opm-tests a bit, and one of the consensus points is to try to gather the different feature tests in the corresponding feature folder.

hakonhagland:
> moving it to the new krnum folder

@alfbr Done!

alfbr:
Thank you! Merging.

PR #845, opened 2022-11-04
Aquifers Re-Factoring

OPMUSER:
Combining four aquifer cases into one directory with documentation - WIP


OPMUSER:
Ready for review.

bska:
jenkins build this please

bska (commented):
I appreciate the work here, and I think it should go in, possibly with a couple of additional summary vectors.  That said, I'm on the fence on adding the ECLIPSE reference solutions.  We've usually shied away from doing that in the past even if having that data is useful for comparison purposes.

What do you think, @alfbr?

OPMUSER:
> I  appreciate the work here, and I think it should go in, possibly with a couple of additional summary vectors. That said, I'm on the fence on adding the ECLIPSE reference solutions. We've usually shied away from doing that in the past even if having that data is useful for comparison purposes.

We should definitely have the comparison plots versus the commercial simulator, for a variety of reasons. Having the commercial results is maybe not so important, but if we say we don't keep the commercial results then we need to delete them from the other tests as well, just to be consistent. A halfway solution would be just to keep the commercial simulator's summary output files.

To be clear, we  definitely should have the comparison plots versus the commercial simulator, but it is not strictly necessary to have all the commercial simulator results files.

OPMUSER:
Will rerun the cases using the TUNING keyword to get a better comparison with the commercial simulator, and then update the plots. But for now it off to the Isles :-)

OPMUSER:
@bska This is now ready for review.

PR #851, opened 2022-11-07
Gaslift2

OPMUSER:
This is a refactoring of the GASLIFT test documentation, with reducing the number of plots per case. Now only one plot per well is used; previously it was three plots per well.

Secondly, the cases were rerun with the latest master, and some cases have changed compared to the October runs.

This is ready for review.


bska:
jenkins build this please

bska (approved):
Thanks a lot for doing this work.  I'll merge into master.

PR #855, opened 2022-11-16
Test cases for keyword GRIDUNIT

tskille:
3 model set up with field unit system and 3 models with metric unit system. The same grid input (coord and zorn) is avaliable with grid units metres, feet and cm.


alfbr:
This looks good, great to get grid units properly tested. Merging.

PR #854, opened 2022-11-13
Removing unsupported keywords from test model data decks.

tskille:
Removing keywords that are not supported by Flow. Since these keywords are unsupported, this PR will not impact simulation results. Hence no update of reference solutions should be necessary. The following keywords are removed from test models.

GRIDUNINT, OPTIONS, PIMULTAB, RESVNUM, COORDSYS and GCONTOL, WPITAB, GECON and GRUPNET.

PR #862, opened 2022-11-29
 Removing keywords that are not supported by OPM Flow

OPMUSER:
Removing keywords that are not supported by OPM Flow, and that would impact results:  GRIDUNIT and OPTIONS keywords.


bska:
For what it's worth, we do support `GRIDUNIT`.  PR #855 even added a number of test cases to verify that the feature works.

OPMUSER:
Arrgghhhhhhhhhhhh. I was just following https://github.com/OPM/opm-tests/pull/854 to make sure that the re-factored cases are consistent.

Never mind, no harm done anyway.

OPMUSER:
Closing as redundant.

PR #866, opened 2022-12-07
Wpimult Update

OPMUSER:
Updated documentation using Markdown and removed ODP file. Reran cases on five day time steps for better comparison with E100 and removed RSM output, plus various other cleanups.


OPMUSER:
Ready for review

bska (approved):
While I think we should generally not add ECLIPSE simulation results to OPM Tests, I'm going to merge this for the other updates.  We may consider scrubbing ECLIPSE output at a later time.

PR #867, opened 2022-12-07
add co2store + gas + water case

totto82:
This add a test case for CO2STORE in combination with GAS and WATER

The should be added as a regression test once https://github.com/OPM/opm-simulators/pull/4292 and friends is merged.


alfbr:
Thanks for the contribution. This is a new data set, and a welcome addition. Merging.

PR #868, opened 2022-12-07
removing last four months in mod4_no_action.sch

GitPaean:
totally chaotic simulation, there is no real meaning to compare.

The simulation output looks like something like follows and resulting well curves are volatile. 
```
Starting time step 30, stepsize 0.0983713 days, at day 388.673/399, date = 21-Aug-2020
 well WIL1 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL1 gets SHUT during iteration 
 well WIL2 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL2 gets REVIVED during iteration 
    Oscillating behavior detected: Relaxation set to 0.900000
 well WIL1 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL2 gets SHUT during iteration 
 well WIL2 gets REVIVED during iteration 
    Oscillating behavior detected: Relaxation set to 0.800000
 well WIL1 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
    Oscillating behavior detected: Relaxation set to 0.700000
    The control model for well WIL2 is oscillating
    We don't allow for more than 3 switches. The control is kept at BHP
    The control model for well WIL1 is oscillating
    We don't allow for more than 3 switches. The control is kept at GRUP
 Newton its=15, linearizations=16 (0.1sec), linear its= 27 (0.0sec)
```

### DIFF
--- a/model4/include/mod4_no_action.sch
+++ b/model4/include/mod4_no_action.sch
@@ -450,16 +450,3 @@ DATES    -- 51
 DATES    -- 52
  1 AUG 2020 /
 /
-
-DATES    -- 53
- 1 SEP 2020 /
-/
-
-DATES    -- 54
- 1 OCT 2020 /
-/
-
-DATES    -- 55
- 1 NOV 2020 /
-/
-


OPMUSER:
@GitPaean Looking at this now.

OPMUSER:
Strange that this an issue with water injectors, usually the issues we have is with the producers. Both WIL1 and WIL2 are the cause of the instabilities. For the WIL1, the 2022-10 release has this:

image

Now comparing this with my reference solution (2021-11-03)

image

Well WIL2 shows the same behavior in both cases.

I noticed that the master reference results all have different dates for the same run (same with my reference results) - why is that? I would of expected all the files would have the same date.

Now the decks have a TUNING keyword setting the maximum time step size to one day, that is what is used in the 2022-10 run, whereas, the reference solution appears to not to use the TUNING keyword. So re-running without the TUNING keyword, we get this:
 
image

Looks better, but we still get the oscillations

Comparing the WBHP pressures for the 2022-10 and the no TUNING case we get:

image

Now comparing this with my reference solution (2021-11-03)

image

There is a maximum WBHP constraint of 200 barsa on all injectors, so it looks like there is an issue with the water injectors, not switching to BHP control correctly, instead we get these messages:

```
Starting time step 114, stepsize 0.190613 days, at day 409.979/429, date = 11-Sep-2020
 well WIL1 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL1 gets SHUT during iteration 
 well WIL2 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL2 gets REVIVED during iteration 
 well WIL1 gets SHUT during iteration 
 well WIL2 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL2 gets REVIVED during iteration 
 well WIL1 gets SHUT during iteration 
 well WIL2 gets SHUT during iteration 
 well WIL1 gets REVIVED during iteration 
 well WIL2 gets REVIVED during iteration 
    The control model for well WIL1 is oscillating
    We don't allow for more than 3 switches. The control is kept at GRUP
    The control model for well WIL2 is oscillating
    We don't allow for more than 3 switches. The control is kept at GRUP
 Newton its=14, linearizations=15 (0.1sec), linear its= 18 (0.1sec)      

```
The wells should switch to BHP control and inject at the rate governed by the BHP, not shutting -in or changing to group control. It looks like the logic being used is for producers, where the BHP is a minimum, whereas, for injectors it is a maximum.

Based on the above we have a regression, and thus I don't think we can just edit out this issue by shortening the SCHEDULE section., but instead fix the problem.

OPMUSER:
>What do you mean by the reference solution here? Is it obtained by OPM-flow? With TUNING or not?

The reference solution is a OPM Flow run and does not have the TUNING keyword active, as it was run with `enable-tuning=false` - the default. So I reran the case with 2022-10 with the TUNING keyword commented out. As I mentioned, it looks better but we still see the same symptoms.

GitPaean:
I see. Your point is that from 2021-11-03 to 22-10 release, there some regression happened.  The result from 2021-11-03 look indeed good.  Then I agree with your point about fixing the regression happened after 2021-11-03. 

Closing this PR now.

PR #872, opened 2022-12-11
test cases with pinch all and multz-

tskille:
This PR includes two new test cases with PINCH item 5 = ALL and keyword MULTZ-. 

- PINCH_MULTZ-_ALL.DATA
- PINCH_MULTZ-_ALL_BARRIER.DATA

The second model (multz- barrier) is currently not supported with Flow, hence the simulator will need a fix before this can be put under regression testing.

These two new test cases are similar to existing cases [PINCH_MULTZ_ALL](https://github.com/OPM/opm-tests/blob/master/pinch/PINCH_MULTZ_ALL.DATA) and [PINCH_MULTZ_ALL_BARRIER](https://github.com/OPM/opm-tests/blob/master/pinch/PINCH_MULTZ_ALL_BARRIER.DATA). The only difference is that the new test cases are using MULTZ- instead of MULTZ.

### DIFF
--- a/pinch/PINCH_MULTZ-_ALL.DATA
+++ b/pinch/PINCH_MULTZ-_ALL.DATA
@@ -1,0 +1,267 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2020 Equinor
+
+-- test case is testing vertical communication set up with
+-- PINCH (item 5 = All) and MULTZ- = 1.0
+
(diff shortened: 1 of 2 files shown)


OPMUSER:
@tskille The two new cases look fine. 

However, when I run the new cases and compare with MULTZ+ versions, there is a discrepancy between the MULT+/MULT- ALL_BARRIER versions, as shown below:

Oil_phase_Pressure

I would have thought that both the ALL_BARRIER cases (bottom figure) would be the same, like the two ALL cases are the same (top figure). I'm I missing something? Does E100 show the same behavior.

tskille:
@OPMUSER , this is correct. @blattms is working on a fix for this.

OPMUSER:
Okay then, feel free to merge so that we have the test cases in master.

OPMUSER:
Closing as fixed.

PR #873, opened 2022-12-20
adding a few cases for WINJMULT keyword

GitPaean:
preparing for regressing tests.


OPMUSER:
@GitPaean some comments here regarding the naming conventions etc. Long file filenames should be avoided for DATA files, because it can be challenging plotting the results when comparing against other cases. So here we are testing the WINJMULT keyword, so why not name the cases `WINJMULT-01,` or `WINJMULT-STW `(for standard wells) and `WINJMULT-MSW `(for multi segment wells) , or something similar. And for a second case used `WINJMULT-STW-01`, etc.

The include files don't need to have a restriction on the filename length, but some suggestions on naming conventions will help other users in understanding what is included in the file:

- Extension should be *.inc so we know it is an include file.
- Use the model case name as a prefix followed by either the keyword or the data type. If different models use the same include files then use the existing include files.

Some examples: 

- So instead of `eqlnum_basemod1.inc` use `WINJMULT-EQLNUM.inc`, I have a preference for upper case characters, but lower case is okay as well. Using the same model (WINJMULT) but with a different EQLNUM array use `WINJMULT-EQLNUM-01.inc`. For a different model (WINJMULT01 or WINJMULT-01) use `WINJMULT01-EQLNUM.inc`

- Instead of `history_basemod1.sch` use something like this  `WINJMULT-STW-SCHD.inc`, or `WINJMULT-STW-SCHD-HIST.inc` 

- Instead of `pvt_basemod1.pvt`  use `WINJMULT-PVT.inc`

This results in a model's include files to be sensibility sorted in alphabetic order.

For verbose data sets in the include file use the NOECHO keyword just before the data, and the ECHO at the end of the file. This makes no difference to OPM Flow, but it will limit the output to the print file for the commercial simulator. So for example:

```
NOECHO

SWATINIT
   0.11596480E+00   0.14962322E+00   0.14084139E+00   0.14851579E+00
   0.21403949E+00   0.10000000E+01   0.76056585E-01   0.76191418E-01
   0.18241443E+00   0.22560194E+00   0.24176513E+00   0.10000000E+01
   0.11189824E+00   0.15942319E+00   0.15003146E+00   0.14386076E+00
   0.30019617E+00   0.10000000E+01   0.11247717E+00   0.86801238E-01
....................................................
/

ECHO
```

For the SUMMARY specification don't specify all the required vectors the way you currently have, but use the ALL keyword, to get the majority and then add the missing ones, that is:

```
-- =================================================================================================================================
-- 
-- SUMMARY SECTION 
-- 
-- =================================================================================================================================

SUMMARY
--
--       EXPORT STANDARD SUMMARY VARIABLE VECTORS TO FILE
--
ALL
-- ---------------------------------------------------------------------------------------------------------------------------------
-- MULTI-SEGEMENT WELL VARIABLES                                                      
-- ---------------------------------------------------------------------------------------------------------------------------------
-- 
--       WELL MULTI-SEGMENT GAS FLOW VARIABLES 
--
SGFR 
'PROD01' /
/
SGFRF 
'PROD01' /
/
SGFRS 
'PROD01' /
/
-- SGFTA 
-- /
-- SGFT 
-- /
-- SGFV 
-- /
-- SGHF 
-- /
-- 
--       WELL MULTI-SEGMENT OIL FLOW VARIABLES 
--
-- SOFTA
-- /
SOFR
'PROD01' /
/
SOFRF
'PROD01' /
/
SOFRS
'PROD01' /
/ 
-- SOFT 
-- /
-- SOFV
-- /
-- SOHF 
-- /
-- 
--       WELL MULTI-SEGMENT WATER FLOW VARIABLES 
--
SWFR 
'PROD01' /
/
-- SWFT 
-- /
-- SWFTA 
-- /
-- SWFV 
-- /
-- SWHF 
-- /
-- 
--       WELL MULTI-SEGMENT RATIO VARIABLES 
--
SGOR 
'PROD01' /
/
SOGR 
'PROD01' /
/
SWCT 
'PROD01' /
/
SWGR 
'PROD01' /
/
-- 
--       WELL MULTI-SEGMENT PRESSURE VARIABLES 
--
SPR
'PROD01' /
/ 
SPRD
'PROD01' /
/
SPRDF
'PROD01' /
/
SPRDH
'PROD01' /
/
SPRDA
'PROD01' /
/
-- SPRDM 
-- /
```
Notice how I commented out the vectors that we currently don't support. We should for multi-segment wells always write out the supported vectors for this well type for checking, IMHO. Note also that liquid volumes are not included with the ALL keyword, so you may have to enter them, if required. Finally, @bska has forbidden the writing out of RSM files, but I think most developers are already aware of this. 

Now I realize that having short file names does have the disadvantage of not fully describing the cases, but these type of long file file name are really only useful to the original author. To overcome, this we should have a Markdown file describing the cases and results. I will do this as part of the refactoring exercise, as it will make life easier for one and all and give confidence to users that the OPM Flow can duplicate the commercial simulator results.

Hopefully the above was useful.

GitPaean:
Thanks for the comments. It was the very beginning of the PR. I will try incorporate your comments when revising it.

OPMUSER:
@GitPaean very well laid out and with good naming conventions. 

I have a few comments, some of which are nitpicks.

- Update Copyright in all files to 2023 instead of 2018

- In *.DATA files, consider changing the description format from:

```
-- ------------------------------------------------------------------------
-- --This is a model based on model 1, WINJMULT keyword is specified ------
-- --with CIRR and CREV mode for two connections respectively.  -----------
-- ------------------------------------------------------------------------
```

to:

```
-- This is a model based on model 1, WINJMULT keyword is specified 
-- with CIRR and CREV mode for two connections respectively.  
```

- Consider changing the TITLE to the case name.

- Remove:

```
-- Turns off echoing of all the input files to the print file.
NOECHO
```

Although this makes no difference to OPM Flow, it does for the commercial simulator, and we want to see what records are being processed. 

- Consider removing EXTRAPMS as it is not supported in OPM Flow.

- My understanding is that only the first record of the TUNING keyword is supported by OPM Flow, so consider changing:

```
TUNING
 1  5  7*  1 /
 /
 12  1  50  1  50  50   /

```
to

```
TUNING
 1  5  7*  1 /
 /
 /
```
Also consider removing 

```
-- Stack size for linear solver
-- No point in nstack > litmax in TUNING
NSTACK
 50 /
```

If you change TUNING.

- In all include files change:

 `-- This file is one of the include files for BASE_MODEL_1.DATA.`

to:

` -- This file is one of the include files for WINJMULT tests.`

Or something similar.

- In the WINJMULT-SUMMARY.inc file,  

    • Remove all the history variables, as this not a history match run.
    • Why so many variables, individually declared? I would have thought the ALL keyword would cover everything, plus WPI, CPI, WBP and WBP4. Note, liquid rates are not covered by the ALL keyword.
    • If a variable is required, but is currently not supported, then consider commenting out in the deck.

- Not a big deal, but why PROD-2 and INJ-6, instead of PROD-1 and INJ-1, or OP-01 and WI-01?

I ran the tests under E100, as I'm Git challenged to run the OPM Flow version of your PR, and I notice that there is not much difference between the three cases.

image

image

Thus, under regression testing, it might be difficult to distinguish a regression failure, since the results are similar. I leave this up to you, but what about setting all connections to CIRR, then CREV, then WREV, and then a mixed case.

Hopefully, the above is of some use.

GitPaean:
> CPI, WBP and WBP4 

The above keywords are not covered by ALL

GitPaean:
The difference is significant enough for regression test. Not totally sure whether to make a mixed case, since not sure whether it is a common practice.

GitPaean:
@OPMUSER , can this PR goes in? It has been here for long and I am coming back to work on this.

bska:
> jenkins build this please

Would you mind rebasing this onto current master as well, please?

GitPaean:
jenkins build this please

bska (approved):

PR #875, opened 2022-12-21
MSW Re-factoring

OPMUSER:
Re-factored two MSW cases and added additional tests for checking for missing data. The additional tests are not meant for regression testing.

There is an issue the MSW-2D-HZ test as it failed in the commercial when processing the third set of re-completions at 2019-12-14.  See https://github.com/OPM/opm-tests/pull/873 for notes.


bska:
jenkins build this please

GitPaean:
> See https://github.com/OPM/opm-tests/pull/873 for notes.

I think it should refer to a different situation?

PR #878, opened 2023-01-05
PINCH Case Documentation and Refactoring

OPMUSER:
Rename PINCH Cases

Note Five_PINCH_NOGAP and Five_PINCH_NOGAP1 are the same cases so drop Five_PINCH_NOGAP.


OPMUSER:
Ready for review.

PR #879, opened 2023-01-05
Test case for constant aquifer flux

tskille:
New test case with constant aquifer flux, keyword AQUFLUX. This is currently not supported in OPM flow.


OPMUSER:
@tskille Should we update the copyright, as this is a new test?

> -- Copyright (C) 2015 Statoil
> -- Copyright (C) 2015 SINTEF ICT, Applied Mathematics

The SWAT, SGAS, and SOIL options are E300, do you think we should remove them? The data is always written out for E100 runs and the run is oil-water.

> RPTRST
> BASIC=2 KRO KRW KRG SWAT PRES SGAS SOIL /  

Other than that - looks good, and the model runs fine in the commercial simulator.

Merge when ready.

tskille:
Copyright header is updated.

I started out with a copy of test case AQUFET-01.DATA. when creating this case. I have not review this test case before using this. 

> The SWAT, SGAS, and SOIL options are E300, do you think we should remove them?

I'm happy to remove these and this is done in the updated commit.

OPMUSER:
Looks good - merging

PR #881, opened 2023-01-09
GPMAINT Test Case - WIP

OPMUSER:
Currently we don't have any GPMAINT regression tests, which is surprising. The initial tests will check both the use of FIPNUM and FIPxxx usage, and will use MODEL04 as the basis.


OPMUSER:
None of the three tests cases run in OPM Flow. I would therefore suggest we merge these now, so that they can be used for testing and debugging. Once these cases are working, I'll update the documentation and add additional  more complicated tests

bska:
> we don't have any GPMAINT regression tests

True.

> I would suggest we merge these now, so that they can be used for testing and debugging.

@totto82 was the one who did most of the work on GPMAINT.  Let's ask him what the current level of support is what, if any, the future plans are for that keyword.

totto82:
I most admit I don't remember all the details. I think the basic usage should work (or at least worked). Torbjørn made a test case, and I also think he tested it on a field model, but I dont remember if he was happy with the results or not. I can try to run the test cases in this PR and see if it helps my memory.

totto82:
The code crash here 
https://github.com/OPM/opm-simulators/blob/c5be1226d7e470c7652f470abb4e79aa6d558395/opm/simulators/wells/WellGroupControls.cpp#L276

due to missing injection control. The simulator expect a GCONINJE when GPMAINT is used.

totto82:
I think the solution is to create a minimum groupInjectionProperties object from GPMAINT i.e by setting the control model and phase based on the GPMAINT item 2.

tskille:
The usage of GPMAINT that was tested in september 2021 is the following 

```
GPMAINT
 'RES' 'PROD' 1 1* 270 3 7 /
/
```

are we supporting other that PROD in item 2 of GPMAINT ? 

We should update [the partly supported file](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp).  Currently there is now entry for this keyword

OPMUSER:
But E100 runs okay. Will add three more tests to check if OPM Flow will
work with the additional keywords, but I think these are optional.

The tests were developed for a client's Shortcut report.

totto82:
> But E100 runs okay. Will add three more tests to check if OPM Flow will
> work with the additional keywords, but I think these are optional

I already made a fix that make it possible to run the case without GCONINJE. I will clean it up and make a PR. I dont think it is necessary to add additional tests for the combined case as OPM Flow should run without GCONINJE.

totto82:
I pushed a fix. With the fix, case 1 and 2 runs. Have not looked at the results yet. Case 3 faults due to issue with the FIPNUM regions or rather FIPLAYER and FIPAREAS I guess. Are we suppose to support  FIPXXX for any XXX? Never seen this usage before.

OPMUSER:
>I pushed a fix. With the fix, case 1 and 2 runs. Have not looked at the results yet. Case 3 faults due to issue with the FIPNUM regions or rather FIPLAYER and FIPAREAS I guess. Are we suppose to support FIPXXX for any XXX? Never seen this usage before.

@totto82 That was fast with the fix. :1st_place_medal: 

I don't think the issue is with FIPXXXX as FIPNUM and FIPLAYER both work and both use 2 as the region number. Perhaps the problem is with FIPAREAS using 6, when perhaps the maximum value is considered to be only 2. 

Yes we are suppose to support the FIPXXXXX and Case 2 shows that we do.

I'll download the client's model to check the formulation for the model failure

totto82:
> I don't think the issue is with FIPXXXX as FIPNUM and FIPLAYER both work and both use 2 as the region number. Perhaps the problem is with FIPAREAS using 6, when perhaps the maximum value is considered to be only 2.

When I look at the FIP value used in the code it only contains 1 and 2. i.e. it does not take into account FIPAREAS. 

It quarry the fips like this

`const auto& fipnum = fp.get_int("FIPNUM");`

so at least for gpmaint it does not handle any other FIP*  type input than FIPNUM.

OPMUSER:
Okay, then I guess we have to fix this as the client's model has a total of seven FIPXXXXX regions, although only one is used with the GPMAINT keyword.

One question, do we support all GPMAINT(GRPCNTL) options:

1. NONE
2. GINJ
3. GINS
4. OINJ
5. OINS
6. PROD
7. WINJ
8. WINS

totto82:
> One question, do we support all GPMAINT(GRPCNTL) options:

Yes, but as @tskille said, only the PROD is really tested. 

>Okay, then I guess we have to fix this as the client's model has a total of seven FIPXXXXX regions, although only one is used with the GPMAINT keyword.

Good if you have a work-around. I dont think it will be hard to fix this in the simulator. I can make a test and report back.

OPMUSER:
Sorry, I don't have a work around for this unfortunately.

I'll generate some more tests based on using FIPNUM, and then duplicate then for using FIPXXX, then we can select which ones we would like to add for regression testing. Is that okay?

totto82:
I have updated the code with a fix for FIP*. Now all cases run. (still not checked the results)

OPMUSER:
Closing and will open as a new PR

PR #885, opened 2023-01-18
GPMAINT Test Cases

OPMUSER:
A total of ten GPMAINT test cases, note that only one actually runs in OPM Flow, all the other cases fail.

This is ready for review and merging.


bska (commented):
I haven't looked at this in any detail yet, but I notice that there's a .ZIP file named
```
gpmaint/plots/ECL/backup.zip
```
added as part of this PR.  Is that intentional?

OPMUSER:
> I haven't looked at this in any detail yet, but I notice that there's a .ZIP file named
> 
> gpmaint/plots/ECL/backup.zip
> 
> added as part of this PR. Is that intentional?
> 
No that is an error, I'll delete that.

OPMUSER:
Done.

blattms:
@OPMUSER this will keep the zip in the public history. Probably not what we want. You will need to rebase and fixup the "dbca6a72f Plot documentation" commit with the last one using git rebase -i dbca6a72f`.:
In the file that opens, reorder the lines such that the last commit is below dbca6a72f and change the front of the line to fixup, Then save and continue.

blattms:
Successor of #881

OPMUSER:
I have update the results based on the latest master, and results are significantly better -all  the cases actually run. However, some cases are a not matching the commercial simulator as shown in the` GPMAINT.md` file. Secondly, there appears to be an issue with OPM Flow not writing out the FIPNUM and FIP region data, in most of theses tests.

I think this should go in as is, so that we have a benchmark.

OPMUSER:
> My preference is that this is not the right place to include plots from Eclipse. For developers who need those plots when implementing features, they can be shared directly instead, e.g., via e-mail. For those of us who clone this repository, the files add significant disk space and download time. They will be part of the history after we remove them. 

The plots are all either jpeg or png files, so they don't take up much space, about 3 MB in the plot directory. So I don't think space requirements and download times are a concern. The idea of displaying the results like this is to give users confidence that OPM Flow results are comparable with the commercial simulator. Secondly, the intent is to only update the plots if the functionality of the keyword has been improved. I also think it is an improvement if keep a graphical history, as there has been cases in which we have had a degradation in results that slipped by without notice.

> The other concern is more a question, and that is the sheer number of cases here is large. It is ten cases for one keyword. If we add them all under regression testing, this may add up quite a bit. Have you looked at the run time?

The intent was to test all functionality of the keyword, but only those tests that align with the commercial simulator should be considered for regression testing, so cases 1,2 ,3, 6 and 7. However, tests 1, 2 and 3 basically test the same functionality but using different FIP array options, so I think we only need to pick one of these for regression testing, perhaps GPMAINT-01. Similarly, for test 6 and 7, we should perhaps use GPMAINT-06 for regression testing.

In terms of run time, on my very old Thinkpad P50 runiing OPM FLow in a VM (now WSL), the runs take about 300 to 400 secs. I think on more modern hardware his will be significantly less.  Perhaps you can test the GPMAINT-01 and GPMAINT-06 cases time on your machine, in order to confirm that the run times are acceptable.

alfbr:
> The plots are all either jpeg or png files, so they don't take up much space, about 3 MB in the plot directory. So I don't think space requirements and download times are a concern.

Fair point. They are also 43 files, so across all tests it adds up a bit. Still, the binary reference solutions may make it a trivial addition to download times. The download times and disk usage is painfully large already though, so I do think we need to start thinking about that. However, I still believe that these plots should be removed or replaced with solutions from Flow once Flow is confirmed to give correct results.

> The idea of displaying the results like this is to give users confidence that OPM Flow results are comparable with the commercial simulator.

I still do not believe github is the right place for that.

>Secondly, the intent is to only update the plots if the functionality of the keyword has been improved. I also think it is an improvement if keep a graphical history, as there has been cases in which we have had a degradation in results that slipped by without notice.

We have already had broad discussions with developers on how to make the update-data workflow easier to do, reducing the risk for introducing new degraded results. It seems we need to align that with the steps made here, but this should not stop the pull request. This is a step in the right direction, let's pick up the discussion again when we find time.

> The intent was to test all functionality of the keyword, but only those tests that align with the commercial simulator should be considered for regression testing, so cases 1,2 ,3, 6 and 7. However, tests 1, 2 and 3 basically test the same functionality but using different FIP array options, so I think we only need to pick one of these for regression testing, perhaps GPMAINT-01. Similarly, for test 6 and 7, we should perhaps use GPMAINT-06 for regression testing.

That sounds good to me. If you are willing to unify 1,2 and 3 that is a nice reduction, maybe you could spread the various FIP options on the remaining tests? Unifying 6 and 7 would also help. Please let me know what you think. If you do not want to reduce the number of tests, we can merge as is, if not, I will wait for an update.
 
> In terms of run time, on my very old Thinkpad P50 runiing OPM FLow in a VM (now WSL), the runs take about 300 to 400 secs. I think on more modern hardware his will be significantly less. Perhaps you can test the GPMAINT-01 and GPMAINT-06 cases time on your machine, in order to confirm that the run times are acceptable.

The run times are a bit on the high side. Ideally, a test should run within a couple seconds. However, this is wishful thinking. In reality it often requires much work to reduce the run time without degrading the usefulness of the test. Testing of group functionality is difficult in this respect. Hence, we need to be practical, a long running test is typically better than no test. I will merge this PR with the current run time. However, if you do see good ways of reducing the run time (e.g., by compacting the time span in the SCHEDULE section). that would be a nice improvement before putting them under regression testing.

OPMUSER:
> That sounds good to me. If you are willing to unify 1,2 and 3 that is a nice reduction, maybe you could spread the various FIP options on the remaining tests? Unifying 6 and 7 would also help. Please let me know what you think. If you do not want to reduce the number of tests, we can merge as is, if not, I will wait for an update.

To be clear, I think we should keep all tests, but only put GPMAINT-01 and GPMAINT-06 under regression testing. The tests are designed to test a limited number of options per case. The reason for this, is that when combining multiple options, it is unclear which option is causing the divergence. Note also that in some cases not all the OPM Flow summary vectors are being written out. For example in GPMAINT-01, FIPNUM region 2 oil, gas, water volumes are missing, although the RPR is written out, this is a bug that needs to be addressed.  

> However, if you do see good ways of reducing the run time (e.g., by compacting the time span in the SCHEDULE section). that would be a nice improvement before putting them under regression testing.

All the tests run to the same year, which for some tests is important, whereas, for one or two, it is not. I would therefore prefer to keep them all the same, so to compare against one another in the future. 

> The download times and disk usage is painfully large already though, so I do think we need to start thinking about that. 

Currently, we have many repositories that are depreciated, perhaps we should archive them and take them offline.

alfbr:
Thanks for the discussion, in it goes.

PR #888, opened 2023-02-08
Test case using JFUNC keyword

tskille:
Keyword used to scale oil-water capillary pressure

Notice that this JFUNC did work correctly on this case on 2023-01-30, but was broken 2023-01-31.
Average water saturation in the model initialize with Eclipse (and Flow before 2023-01-31) is 16.4%. 

Average water saturation using flow build with master after 2023-01-31 is significantly off (10.0%).


alfbr:
Thanks for the contribution, highly appreciated! Merging.

PR #889, opened 2023-02-09
Test case with ROCKOPTS and default ROCK records

tskille:
Notice that this test case is currently not supported by OPM Flow.

This data deck includes the following ROCK keyword in PROPS section

```
ROCK
-- REF. PRES   COMPRESSIBILITY
   280.000        5.6E-5 /
/  -- default, copy of previous record
```

If this is modified to the equivalent 

```
ROCK
-- REF. PRES   COMPRESSIBILITY
   280.000        5.6E-5 /
   280.000        5.6E-5 /
```

flow will handle this correctly.


bska (approved):
Thanks a lot.  I'll merge into master and then we'll get started on solving the underlying problem.

PR #891, opened 2023-02-13
Test case with ACTIONX and WCONHIST

tskille:
This test model is not supported by OPM but hopefully it should not be hard to fix this.

### DIFF
--- a/actionx/ACTIONX_WCONHIST.DATA
+++ b/actionx/ACTIONX_WCONHIST.DATA
@@ -1,0 +1,424 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2023 Equinor
+
+-- This model is built from the base case for model2. This case has standard wells, 1 producer and 1 injector.
+-- There is two actionx keywords in the deck. The forst action is using WCONHIST to change the control rate for an
+-- existing producer. The second action is used to change an injetor to a producer.
+--
+-- =================================================================================================================================
+--
+-- RUNSPEC SECTION
+--
+-- =================================================================================================================================
+RUNSPEC
+--
+--       DEFINE THE TITLE FOR THE RUN
+--
+TITLE
+OPM_TESTS: ACTIONX_WCONHIST
+--
+--       DEFINE THE START DATE FOR THE RUN
+--
+START
+         31 'DEC' 2017                                                         /
+--
+--       SWITCH NO SIMULATION MODE FOR DATA CHECKING COMMENT OUT TO RUN THE MODEL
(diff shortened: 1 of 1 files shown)


alfbr:
Thank you! Great to see the runtime under one second and improved test coverage of ACTIONX. Merging.

OPMUSER:
If you add support for the WCONHIST keyword within an ACTIONX block, via /opm-common/src/opm/input/eclipse/Schedule/Action/ActionX.cpp, then the case fails because:

```
Report step  2/4 at day 5/20, date = 05-Jan-2018

Starting time step 0, stepsize 5 days, at day 5/10, date = 05-Jan-2018
Restart file written for report step 2/4, date = 05-Jan-2018 00:00:00

Warning: Keyword 'WELLS' is unhandled for output to file.
Action WHWI01 triggered at 10-Jan-2018 00:00:00 (report interval 2 to 3).
Program threw an exception: Trying to get injection data from a producer
Failed: 6648
OPM Flow Process 6648 Has Failed
Process Complete (0) 
```
This is because WI01 is first declared as an water injector, and then in the ACTIONX block it is re-specified as a oil producer using WELSPECS, with the oil production rate set via WCONHIST. I think this is a schedule issue, that is the well has not changed status went looking up the SUMMARY data.

alfbr:
> If you add support for the WCONHIST keyword within an ACTIONX block, via /opm-common/src/opm/input/eclipse/Schedule/Action/ActionX.cpp, then the case fails because:
> 
> ```
> Report step  2/4 at day 5/20, date = 05-Jan-2018
> Starting time step 0, stepsize 5 days, at day 5/10, date = 05-Jan-2018
> Restart file written for report step 2/4, date = 05-Jan-2018 00:00:00
> 
> Warning: Keyword 'WELLS' is unhandled for output to file.
> Action WHWI01 triggered at 10-Jan-2018 00:00:00 (report interval 2 to 3).
> Program threw an exception: Trying to get injection data from a producer
> Failed: 6648
> OPM Flow Process 6648 Has Failed
> Process Complete (0) 
> ```
>
> This is because WI01 is first declared as an water injector, and then in the ACTIONX block it is re-specified as a oil producer using WELSPECS, with the oil production rate set via WCONHIST.

Yes, we do not currently support combinin WELSPECS with WCONHIST in an ACTIONX block. However, note that the test has two ACTIONX blocks. The first only tests WCONHIST, and works well. 

> I think this is a schedule issue, that is the well has not changed status went looking up the SUMMARY data.

Not sure what you mean by schedule issue, but I believe you are right that WELSPECS is not handled correctly in an ACTIONX block. This should be fixed before the deck is put under testing.

OPMUSER:
> Not sure what you mean by schedule issue, but I believe you are right that WELSPECS is not handled correctly in an ACTIONX block. This should be fixed before the deck is put under testing.

What I mean by "schedule issue", is a really a possible timing issue. An ACTIONX block is supposed to be executed at the next time step, so until then the WI01 is still a water injector, creating a possible conflict with the current state of the well.

PR #893, opened 2023-02-13
Test case with ACTIONX and WCONINJH

tskille:
Currently not supported by OPM Flow

### DIFF
--- a/actionx/ACTIONX_WCONINJH.DATA
+++ b/actionx/ACTIONX_WCONINJH.DATA
@@ -1,0 +1,431 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2023 Equinor
+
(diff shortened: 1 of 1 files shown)


alfbr:
Thank you, greatly appreciated! Merging.

OPMUSER:
If you add support for the WCONINJH keyword within an ACTIONX block, via /opm-common/src/opm/input/eclipse/Schedule/Action/ActionX.cpp, then the case fails because:

```
Report step  2/4 at day 5/20, date = 05-Jan-2018

Starting time step 0, stepsize 5 days, at day 5/10, date = 05-Jan-2018
Restart file written for report step 2/4, date = 05-Jan-2018 00:00:00

Warning: Keyword 'WELLS' is unhandled for output to file.
Action WIWI02 triggered at 10-Jan-2018 00:00:00 (report interval 2 to 3).
Program threw an exception: Trying to get production data from an injector
Failed: 6559
OPM Flow Process 6559 Has Failed
Process Complete (0) 
```
This is because OP02 is first declared as an oil producer, and then in the ACTIONX block it is re-specified as a water injector using WELSPECS, with water injection rate set via WCONINJH. I think this is a schedule issue, that is the well has not changed status went looking up the SUMMARY data.

This works as planned in the commercial simulator.

PR #894, opened 2023-02-14
Comment out NONNC in SPE10 Cases

OPMUSER:
Comment out the NONNC keyword in SPE10 cases, as the keyword is not supported, and results in the cases not running.


bska:
jenkins build this please

bska (approved):

PR #898, opened 2023-02-17
OPERATER Keyword: Remove RUNSUM keyword from All Tests

OPMUSER:
Remove RUNSUM keyword from all tests in preparation for future regression testing.


bska (approved):

PR #903, opened 2023-02-20
Add CO2STORE + THERMAL case

goncalvesmachadoc:


totto82:
Thanks for the model. I am able to run this in Flow after removing some compositional keywords and fixing some missing issues. See https://github.com/OPM/opm-common/pull/3410 I also added HEATCRT to the deck. Is the temperature dependence supposed to be neglected? Currently Flow fails if HEATCR is given but not HEATCRT.

bska:
@goncalvesmachadoc: Do you expect to make further changes to this test case?  If not, we should probably try to get the model into the main repository.

goncalvesmachadoc:
@totto82 and I are still planning on modifying it. We need to see which keywords we will comment out, and what we want to do about compositional keywords. Probably keep it, but ignore them. The same for https://github.com/OPM/opm-tests/pull/902 .

bska:
> We need to see which keywords we will comment out, and what we want to do about compositional keywords. Probably keep it, but ignore them. The same for #902 .

Noted.  Thanks a lot for the update!

goncalvesmachadoc:
This case is ready for review. It has a few modifications to be able to run with OPM.

bska (approved):

PR #904, opened 2023-02-20
Test case using WCONINJH and VFP

tskille:
This test case is created as a follow up to PR 4461 in opm-simulators 

   https://github.com/OPM/opm-simulators/pull/4461

Simulation results has been verified and I would very much like to see this test put under regression testing


OPMUSER:
@tskille Consider placing this test in a new vfpprod directory, perhaps even merging the vfpprod_iglr and vfpprod_spe1 directories, rather than the model2 directory. Take a look at the grpcntl directory and the include files, where I have used the model2 files for an example. You can use the same files in your test.

I suggest this because the model2 has got many tests and is thus getting rather confusing, and it would be good to have all the vfp* tests in the same directory.

Also, what about calling the test vfpinj-wconinjh (or vfpinj_wconinjh) instead of inj_vfp_wconinjh?

What do you think?

tskille:
> Consider placing this test in a new vfpprod directory

Notice that this test is testing vfp + wconinjh, hence placing this test in a folder vfpprod is probably not what we want. What about creating a new folder vfp ? 

> perhaps even merging the vfpprod_iglr and vfpprod_spe1

Yes I can merge this but we first have to agree where to put these

OPMUSER:
> Notice that this test is testing vfp + wconinjh, hence placing this test in a folder vfpprod is probably not what we want. What about creating a new folder vfp ?

Yes, I think a new vfp folder is a better idea.

> Yes I can merge this but we first have to agree where to put these

Sounds like we have a plan. If you do the merging etc, then I'll document the cases using markdown, after your PR is merged.

tskille:
the test vfpprod_spe1 has not been moved yet, since this is under regression testing today. Not sure how to do this. @bska should we merge this as is, and then create a separate PR for moving vfpprod_spe1 ?

bska (approved):
I didn't notice before that one of the includes is named "fluxMUN" instead of "fluxNUM" as I would expect.  It obviously doesn't matter since we `INCLUDE` [the correct file](https://github.com/OPM/opm-tests/blob/2f1bb782286f11976a996aec42e67a4608922465/vfp/VFPINJ_WCONINJH-01.DATA#L96-L97) but a bit curious nonetheless.  If you would like to rename the file (MUN :arrow_right: NUM) then go ahead.

Otherwise I'm happy to merge this.

bska:
> the test vfpprod_spe1 has not been moved yet, since this is under regression testing today. [...] should we merge this as is, and then create a separate PR for moving vfpprod_spe1 ?

Yes, I think we should move that case separately since doing that involves synchronising changes among multiple repositories.

PR #911, opened 2023-02-23
Test case with VFP+WCONINJE

tskille:
Case includes 2 producers, 1 gas injector and 1 water injector. Both injectors are initially controlled with maximum surface rate. At the end of the simulation target surface rates are increased and both injectors end up being controlled by maximum tubing head pressure.

### DIFF
--- a/vfp/VFPINJ_WCONINJE-01.DATA
+++ b/vfp/VFPINJ_WCONINJE-01.DATA
@@ -1,0 +1,311 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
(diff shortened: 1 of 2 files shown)


tskille:
Notice that this model is not behaving as expected and needs to be checked by a developer. Flow ends up closing both injectors in the last phase of the simulation.

gas injector INJ1
```
Restart file written for report step 1/3, date = 10-Nov-2018 00:00:00
    Well INJ1 control mode changed from RATE to THP
Well INJ1 will be shut as it can not operate under current reservoir conditions.
```
water injetor INJ2
```
Restart file written for report step 2/3, date = 20-Nov-2018 00:00:00
    Well INJ2 control mode changed from RATE to THP
Well INJ2 will be shut as it can not operate under current reservoir conditions.
```

alfbr:
Thank you, highly appreciated! Merging.

PR #912, opened 2023-02-24
Adding test model for GCONINJE item10=NETV with a gas injector

vkip:
The purpose of this model is to verify that GCONINJE item10=NETV also works with gas injectors present (feature implemented in https://github.com/OPM/opm-simulators/pull/4460 ).


alfbr:
Thank you, highly appreciated! Merging.

PR #916, opened 2023-03-06
Adding test case for PVTNUM as the default ROCK region keyword.

vkip:


alfbr:
Thanks for the test, highly appreciated! This is ready to go in with two small changes.

This seems to be a feature test rather than a test to combine different features in model1. Hence, please copy the DATA file and necessary include files to a separate folder with an intuitive name.

Also, please update the file header with copyright year 2023 and shorten the description of the case to reflect the functionality being tested.

alfbr:
Thank you! Merging.

PR #937, opened 2023-03-30
two new test cases with keyword WGRUPCON

tskille:
* 01-WGRUPCON.DATA
* 02-WGRUPCON.DATA

The first model has WGRUPCON + Item2 = YES (default) and this model is supported by Flow. The second test model uses WGRUPCON + Item2 = NO and this is not supported with the current version of OPM Flow.


bska (approved):
Very good.  Having these cases makes it easier to develop the requisite feature.  I'll merge into master.

PR #942, opened 2023-04-13
Update actionx info

tskille:
Update of markup file for actionx tests


actionx/ACTIONX.md
```diff
@@ -1,53 +1,48 @@
...
+BOX          | ENDBOX, MULTX, MULTY, MULTZ                                         | ACTIONX_BOX        | Prediction | MULT       | No    | Model is supported without ACTIONX keyword
+COMPDAT      | COMPLUMP, WELOPEN                                                   | ACTIONX_COMPDAT    | History    | MODEL02    | Yes   | 
+COMPLUMP     | WELOPEN                                                             | ACTIONX_COMPLUMP   | History    | MODEL02    | Yes   | 
+ECHO         | -                                                                   | -                  | -          | -          | -     | Keyword not supported in deck.
+ENDBOX       | BOX, MULTX, MULTY, MULTZ                                            | See ACTIONX_BOX    | Prediction | MULT       | No    | 
+GCONINJE     | WELOPEN                                                             | ACTIONX_GCONINJE   | Prediction | MODEL02    | No    | !!! model under regression testing, must be removed from regression testing !!!
```
bska, line 10: What does the comment here,

> !!! model under regression testing, must be removed from regression testing !!!

mean?

bska (commented):
In general this looks good to me.  I do wonder about one of the new comments though.

actionx/ACTIONX.md
bska: > Two tests used for regression testing currently have issues and results are not correct. Hence these should be remove from regression testing until these issues are fixed.

Ah, I see.  Thanks a lot for clearing that up!

bska (approved):

PR #945, opened 2023-04-15
Fixed data decks with GCONPROD item 7 not set to RATE

tskille:
Only RATE is supported in GCONPROD item 7. Notice that NONE is default, hence default not supported.

The current behavior of OPM Flow is that the simulator will handle item 7 input in GCONPROD as if this was RATE, independent of what the input actually is. Hence simulation results (all summary vectors and solution data in restart file ) will not changed due to this PR. However, this will modify output to some of the restart arrays (example IGRP). An update of reference solutions are therefor required. 

This is a preparation for a PR in opm-simulators which will stop the simulation if input different that RATE is given.


bska:
jenkins build this please

bska (approved):

PR #950, opened 2023-04-25
Update CO2STORE_GASWAT

goncalvesmachadoc:
Following PR https://github.com/OPM/opm-tests/pull/947


bska:
I think this was also (accidentally?) added directly to the master branch in commit 792bc932.  Would you mind rebasing onto current master?

goncalvesmachadoc:
Oh, my bad. will do.

PR #951, opened 2023-04-25
using RATE for item 7 of GCONPROD in NETWORK-01.DATA

GitPaean:
the default option 'NONE` is not supported by flow. We can not run with the default option after PR https://github.com/OPM/opm-simulators/pull/4593 .


bska:
jenkins build this please

bska (approved):

PR #956, opened 2023-05-08
removing PERFORMA summary keyword from NETWORK-01

GitPaean:
this makes the comparison criteria used in parallel tests not feasible. At the same time, it will be problem for future regression test for this case.


GitPaean:
jenkins build this please

PR #969, opened 2023-05-16
Update markup file for gaslift cases

tskille:


bska (commented):
This looks good to me.  I notice a couple of smaller issues, especially
```
RESTART
 'GASLIFT-14' /
```
in the `SOLUTION` section of `gaslift/GASLIFT-14.DATA` which is unexpected to me.  Was that intended to be in `gaslift/GASLIFT-14_RESTART.DATA` instead?  Other than that, I mostly saw `ECHO` and `NOECHO` keywords.  That's fine by me, but maybe we'd like to filter those out before merging?

Finally, I just learned that GitHub's Markdown engine knows about [mermaid](https://mermaid.js.org/) so if we wanted to we could have mermaid draw our group trees for us, e.g., something along the lines of the example below.

That's for future updates though, and it's not clear to me that it would be a net benefit for offline viewing.

---

```mermaid
flowchart TD
    A[FIELD] --> B{PLAT-A}
    B --> C{M5S}
    B --> D{M5N}
    C --> E{B1}
    C --> F{G1}
    E --> G[B-1H]
    E --> H[B-2H]
    E --> I[B-3H]
    F --> J[G-3H]
    F --> K[G-4H]
    D --> L{C1}
    D --> M{F1}
    L --> N[C-1H]
    L --> O[C-2H]
    M --> P[F-1H]
    M --> Q[F-2H]
```

PR #971, opened 2023-05-24
new test case wih keyword VAPPARS

tskille:
The following test case includes the keyword 
```
VAPPARS
  2.5 0 /
```
in the SOLUTION section of the data deck. 

OPM Flow is currently not supporting this deck.


bska (approved):
Thanks.  I have a vague idea of what's causing the `VAPPARS` problem, but the first step to analysing the underlying problem is getting the model.  I'll merge into master.

PR #985, opened 2023-06-05
New test case with MULTFLT

tskille:
applied on same faults both in GRID and EDIT section. WIth current master OPM Flow is not handling this correctly.


bska (approved):

PR #988, opened 2023-06-12
add L20 to opm-tests

goncalvesmachadoc:


bska (commented):
Would you mind adding the copyright statement to all three include files, please?

goncalvesmachadoc:
This model has 20 layers. I could change the name of include files to make it simpler.

bska:
> > Would you mind adding the copyright statement to all three include files, please?
> 
> Done :)

:+1: Much appreciated!

> This model has 20 layers. I could change the name of include files to make it simpler.

Nah, it's fine I think.  Given that the main `.DATA` file has a descriptive name I'm much less concerned about it.  If the main input file was named `L20.DATA` then I might have requested that it be renamed, but this setup is fine by me.  I'll merge into master.

PR #991, opened 2023-06-14
Two new tests cases for WAGHYSTR

tskille:
This PR includes two new test cases with keyword WAGHYSTR. 

The keyword is not supported yet in OPM flow, but there are two PR in progress 

https://github.com/OPM/opm-common/pull/3542  
https://github.com/OPM/opm-simulators/pull/4583


bska (approved):
Thanks a lot for the contribution.  I'll merge into master.

PR #998, opened 2023-06-26
Speed-up salt precipitation

goncalvesmachadoc:
@bska We can run this now in 45 sec on my machine. Both gas condensation and salt precipitation happens in this simulation timeline.


atgeirr:
Perhaps it is better to put the changes in a separate file instead of modifying the existing one?

bska:
jenkins build this please

bska:

PR #999, opened 2023-06-26
removing performance related SUMMARY for WINJMULT cases

GitPaean:
Currently, the parallel testing only compare the accuracy of the results.  Comparing the performance data is not really sensible.


GitPaean:
jenkins build this please

GitPaean:
It is similar to the PR https://github.com/OPM/opm-tests/pull/956.

bska (approved):
Yeah, that makes sense.  The `MAXDPR` and `MAXDSx` keywords arguably pertain to solution changes, but we don't support them anyway so there's no harm in removing them.  I'll merge into master.

PR #1008, opened 2023-07-05
adding WINJDAM_MSW and WINJDAM_STDW cases

GitPaean:
preparing for the regression test for filtration cake model.


GitPaean:
the include folder is similar to the ones in the folder WINJMULT.  Any suggestions regarding the proper practice here?

GitPaean:
My impression is that we have quite some repeated included files now while we name them differently based on which case is using them.

bska (commented):
> The include folder is similar to the ones in the folder `WINJMULT`. Any suggestions regarding the proper practice here? My impression is that we have quite some repeated included files now while we name them differently based on which case is using them.

Yes, that's true.  At some point we considered refactoring the include file, but the benefit of being able to just copy a case directory elsewhere and run it standalone is too useful to give up for the small savings in diskspace we'd get from just having a single copy of each include file.

On a different note, I notice that the include files use `NOECHO`/`ECHO`.  We don't really support those, so I'd appreciate it if you could remove them.

GitPaean:
Thanks for the comments. NOECHO/ECHO has been removed. And they are ready for review.

bska (approved):

PR #1012, opened 2023-07-14
Prune Data File Execute Bits

bska:
These are confusing on Unix/Linux systems.


bska:
jenkins build this please

alfbr:
Indeed, thanks for fixing! Merging.

PR #1016, opened 2023-07-31
using UDQ value for the case WINJDAM_STDW.DATA

GitPaean:


GitPaean:
At the end, I modified an existing case instead of creating a new case. With this way, we did not lose any testing capacity and also tested the UDQ implementation.

bska:
Merging this now to prepare for merging the upstream PRs.

PR #1030, opened 2023-08-29
Test cases with keyword CSKIN

tskille:
This PR includes a new test case holdin the keyword CSKIN. Notice that this keyword is currently not supported in OPM Flow but work on this is about to start these days


bska (approved):
Thanks.  This looks good to me and I'll merge into master.  I look forward to seeing how the new feature progresses.

PR #1033, opened 2023-08-31
Additional test cases for CSKIN

tskille:
These test cases convers different input for KH (item 8 in COMPDAT) and CSKIN used in combination with keyword WPIMULT and WELPI.


bska (approved):
Thanks.  These interactions clearly need attention when we add support for CSKIN.  I'll merge into master.

PR #1034, opened 2023-09-01
Test cases to verify no change from a completely open valve

vkip:
Ref https://github.com/OPM/opm-simulators/pull/4824


GitPaean:
Thanks for the DATA files. It is sufficient to protect the new development in https://github.com/OPM/opm-simulators/pull/4824, which accounts for the hydrostatic and acceleration pressure drops for the valves. 

I am getting the DATA in now. And I will create a regression test based on [MSW-2D-VERT-01.DATA](https://github.com/OPM/opm-tests/pull/1034/files#diff-5159112502e8a4c2d943bae894dc0f4dfbd1e3b0089a3e351a625962eebbd237).

PR #1037, opened 2023-09-05
Test case with DIRICHLET boundary condition

svenn-t:
Test case for DIRICHLET boundary condition. Made this test case for PR: https://github.com/OPM/opm-simulators/pull/4835


bska (approved):
Looks good to me.  I'll merge into master.

PR #1057, opened 2023-09-28
New test case for Z-transmisibility calculation

tskille:
Test case without keywords PINCH and MINPV. Dimension of grid is 1x1x3 and cell 1,1,2 is fully collapsed (zero thichkess for all corners).

Grid cell 1,1,2 should in this case act as a barrier for vertical flow (tranz = 0.0). OPM Flow is currently not modelling this correct but this will hopefully be fixed soon.


bska (approved):
Thank you for the test case. This looks good to me and I'll merge this into the master branch now.

PR #1061, opened 2023-10-04
Additional test cases for PINCH

tskille:
- Excisting test case renamed from NOMINPV_AND_NOPINCH to T3A_NOPINCH
- 4 additional test cases based on T3A_NOPINCH
- markup file PINCH.md updated


tskille:
all of these  cases shows good result when using 

  https://github.com/OPM/opm-simulators/pull/4901

PR #1062, opened 2023-10-06
Update co2store_gw_dirichlet test case

totto82:
The old test was initially filled with gas. Injecting gas in gas does not make much sense. I also removed the producer to make the boundary conditions more relevant.


totto82:
Any comment on this? I will need to provide an update of the reference data if we decide to merge this change. The current test does not make much sense  IMO.

bska:
> Any comment on this? I will need to provide an update of the reference data if we decide to merge this change.

At the moment, since we're in the lead-up to a new release, we're not merging any PRs which change reference results.  The reference results in OPM-Tests are shared between all branches of OPM Flow and we have traditionally not tagged them with a release number.  Unless we get direct permission from the release manager (@blattms) **and** the CI system manager (@akva2) we will have to wait for the final release before merging these kinds of PRs.

Adding new tests to the regression test suite is fine though.

blattms:
Indeed the test for the release candidate in jenkins uses the master opm-tests. So it would help a lot if it is possible to postpone changes here. Of course if there are high priority fixes that are urgently needed, then this is a different story and we need to find another way. Hopefully this one is not too urgent?

totto82:
Jenkins build this please

PR #1066, opened 2023-10-15
fix illegal spaces in WELSPECS keyword

tskille:
Updated test case with extended `WELSPECS`. Illegal spaces in item number 2 in keyword `WELSPEC` gives error message
```
Error: Problem with keyword WELSPECS
In TEST3_WS.DATA line 314
Illegal space in   B1 when defining WELL/GROUP.
```


bska (approved):

PR #1067, opened 2023-10-17
Add a test case for SWOFLET and SGOFLET based on model2

totto82:
Currently we dont have any tests which uses SWOFLET and SGOFLET. Note that the SWOFLET and SGOFLET input does not match the original tabulated SWOF and SGOF.


totto82:
I really hope to add SWOFLET and SGOFLET to the regression testing. If no further comments. I will selfmerge.

PR #1068, opened 2023-10-18
Adding test case for SWATINIT with a mixed-wet Pc curve

vkip:
Ref https://github.com/OPM/opm-common/pull/3720


bska (approved):

PR #1069, opened 2023-10-24
add test cases for UDT

akva2:
This adds tests cases for UDT provided by @vkip


bska (approved):

PR #1090, opened 2023-11-13
Removing THPRES from network restart tests

vkip:
Ref discussion in https://github.com/OPM/opm-common/pull/3738


GitPaean:
Thanks for the update. I am getting it in now.

PR #1092, opened 2023-11-14
Use CPR to get within thresholds + skip guiderate output

vkip:


vkip:
Ref https://github.com/OPM/opm-simulators/pull/4961 (again)

bska (approved):
Very good.  We may choose to reinstate these summary vectors when the simulator is good enough.

PR #1093, opened 2023-11-22
new test case with GPMAINT

tskille:
This a new test case using GPMAINT with Item2 equal to PROD. 

This functionality used to be supported but there has been a regression some time in May 2023. Hopefully this test case can be used to fix the error and then this should be used for regression testing.


bska (approved):
Thanks a lot for the test case.  I'm sure that this will speed up the process of analysing and fixing the underlying issue.  I'll merge into master.

PR #1098, opened 2023-12-01
Test case for using UDQ in WCONPROD item 12 (ALQ)

vkip:


bska (approved):
Sorry about the delayed response.  This looks good to me and I'll merge into master.

PR #1107, opened 2023-12-15
Add CO2SOL/H2SOL test cases based on SPE1

totto82:
https://github.com/OPM/opm-common/pull/3764


blattms:
Not sure why this has not been merged. Is this still of interest, @totto82?

totto82:
Yes. I would be very happy if this can be merged. I would like to add them in the ctest list.

GitPaean (approved):
Thanks for the update.

PR #1116, opened 2024-01-10
changed: remove CONV from RST output

akva2:
With this implemented, we want to remove the potential source of instability in regression tests.

Sidestream of https://github.com/OPM/opm-simulators/pull/5054


akva2:
jenkins build this please

PR #1123, opened 2024-01-29
adding co2 storage into a depleted gas fields

Negar-Khoshnevis:


alfbr:
Thanks for the contribution! Generally speaking this is ready to go in as is. Being able to run it correctly in Flow is no pre-requisite for submitting new data-sets, and these both represent data-sets we aim to support. Do you want to make any further changes to this pull request?

bska:
> and these both represent data-sets we aim to support.

Right. I don't think we currently support using
```
EQUALS
DX ... /
/
```
as a way of entering grid data.  At least there have been reports (OPM/opm-simulators#4291) that we don't support that method and I don't think we've made any efforts since then to add that kind of support.  Regardless of other issues, you may wish to change that particular aspect of these models.

totto82:
> Regardless of other issues, you may wish to change that particular aspect of these models.
Or keep it. I really want us to support this asap. It is a very convenient way of setting up small test models. Maybe this is the time to implement this.

bska:
> > Regardless of other issues, you may wish to change that particular aspect of these models.
> 
> Or keep it. I really want us to support this asap. It is a very convenient way of setting up small test models. Maybe this is the time to implement this.

The last time I looked into that I got the impression that it was quite a bit of work.

Negar-Khoshnevis:
I will change the parts which are not supported in OPM in the gird section, there are equivalent keywords and easy to change.

bska:
> I will change the parts which are not supported in OPM in the gird section, there are equivalent keywords and easy to change.

Much appreciated.

Negar-Khoshnevis:
> > and these both represent data-sets we aim to support.
> 
> Right. I don't think we currently support using
> 
> ```
> EQUALS
> DX ... /
> /
> ```
> 
> as a way of entering grid data. At least there have been reports ([OPM/opm-simulators#4291](https://github.com/OPM/opm-simulators/issues/4291)) that we don't support that method and I don't think we've made any efforts since then to add that kind of support. Regardless of other issues, you may wish to change that particular aspect of these models.

It is updated with the supported keywords

bska (approved):
Thanks a lot for the updates.  This looks good to me and I'll merge into master.

PR #1126, opened 2024-01-30
Add some reference files for damaris

akva2:
Just a couple of files so we can enable a very basic test.


bska:
Do we have/need a mechanism to update these files if, for instance, the simulator starts emitting more solution vectors?

akva2:
i thought i'd deal with that part once there's more than grid and pressure in there. adding more vectors do nothing to the hdf5 files unless also updating the writing.  it should be a fairly small change, add .h5 to the extensions in the data update script, but it would need testing etc etc so i put that work off until it's more relevant.

atgeirr:
> Do we have/need a mechanism to update these files if, for instance, the simulator starts emitting more solution vectors?

In any case I do not think that is urgent, so I'll merge this and the test that uses it.

PR #1127, opened 2024-01-31
Test case for autochoke option

plgbrts:


vkip:
Thanks, looks good to me - merging.

PR #1129, opened 2024-02-07
Test case for UDQ + Region and ACTIONX + NE

tskille:
This PR includes two new test cases. None of these are currently running in OPM FLow.

UDQ_REG-01.DATA includes region UDQ quanteties (not supported) and ACTIONX_NE.DATA is using the operator != (not equal) and includes a NEXTSTEP inside and ACTIONX block.


bska:
The UDQs in UDQ_REG-01.DATA are, strictly speaking, all defined at the field level (UDQ names `FU*`).  On the other hand their definition does indeed include region level quantities (e.g., `RGPR_RE2 2`) so there is decidedly some work required before the simulator will handle those.

Do we expect to have region level UDQs (`RU*`) as well?

bska (approved):

PR #1131, opened 2024-02-07
Fix number of PVT tables

totto82:
Number of pvt tables and TABDIMS did not match. Also ROCK was duplicated, both in include/PVT-WET-GAS and in the deck.


tskille (approved):
Thanks a lot for spotting this.

PR #1134, opened 2024-02-09
Add test case for FLOWS/FLORES with NNC and BCCON

totto82:
Motivation. 
https://github.com/OPM/opm-models/pull/874


totto82:
The fix is merged. I really want to add this to regression to avoid further issues. Review?

bska:
> I really want to add this to regression to avoid further issues.

Okay, but maybe you can start adding some real unit tests as well?  Regression tests are good as far as they go, but it's a lot easier to catch problems with tests that don't require building and running the full simulator.

totto82:
>Okay, but maybe you can start adding some real unit tests as well? Regression tests are good as far as they go, but it's a lot easier to catch problems with tests that don't require building and running the full simulator.

I dont think it would be easy to make a simple unit test for this without including the full assembly code and the output code, for that we need to include the full simulator anyway. Yes, unit tests are good to have, but regression tests are better than nothing.

PR #1136, opened 2024-03-13
Add test for insert_keyword in PYACTION

lisajulia:
Add test for invoking the insert_keyword - functionality for the keyword WCONPROD

### DIFF
--- a/pyaction/pyaction_wconprod_insert_kw.py
+++ b/pyaction/pyaction_wconprod_insert_kw.py
@@ -1,0 +1,206 @@
+# The python code below does the same as ACT-01 to ACT-05
+import datetime
+
+def run(ecl_state, schedule, report_step, summary_state):
+    print("checking PYACTION\n")
+    start = schedule.start
+    current_time = start + datetime.timedelta(seconds=summary_state.elapsed())
+
+    FU_GOR = summary_state["FU_GOR"]
+    print("FU_GOR {}\n".format(FU_GOR))
+
+    FU_GASFL = summary_state["FU_GASFL"]
+    print("FU_GASFL {}\n".format(FU_GASFL))
+
+    WGOR_OP01 = summary_state.well_var("OP01", "WGOR")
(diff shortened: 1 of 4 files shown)


pyaction/pyaction_wconprod_insert_kw.py
```diff
@@ -0,0 +1,206 @@
...
+        'GI*'    GAS    SHUT   GRUP  1*     1*     1*     1*    1*                     /
+        'WI*'    WAT    SHUT   GRUP  1*     1*     1*     1*    1*                     /
+        /"""
+        schedule.insert_keywords(kw, report_step)
+
+#-- ---------------------------------------------------------------------------------------------------------------------------------
+#-- ACTIONX WCONPROD - BEAM BACK OIL RATE IF GOR >= 150 M3/M3 FOR WELL OP01
+#-- ---------------------------------------------------------------------------------------------------------------------------------
+#ACTIONX
+#        ACT-01    500  10                                  /
```
blattms, line 81: I think the last two parameters might not be fully represented in your script and datafile, yet. We might need multiple PYACTION keywords and python scripts to accomplish this.

From the manual:
ACTNSTEP is a positive integer that defines the number times that the
ACTNAME definition is executed. ACTIONX definitions are activated at
the end of a time step and this parameter is used to set how many time
steps the ACTNAME definition will be invoked. The default value of one
means that the definition will be executed only once. Use a large value, for
example 10,000 for the definition to be executed at every time step.
Noted that the counter only affects successful evaluations; i.e. if
ACTNSTEP is set equal to one (the default), then the simulator will test
the action at the end of every time step until it evaluates to true.

ACTDELTA is a real positive value that stipulates the duration of time that
the conditions defined on the second record to be satisfied before the
ACTIONX action are executed. For example, if ACTDELTA is defaulted
the actions will be executed at the end of the time step for which the
conditions are met. If set to say 30, then a minimum of 30 days will pass
before the actions are executed (assuming field or metric units).

blattms:
jenkins build this please

PR #1138, opened 2024-03-14
Remove actionx_callback

lisajulia:
In https://github.com/OPM/opm-common/pull/3948, I've removed the actionx_callback, this PR removes it in opm-tests as well.

### DIFF
--- a/udq_actionx/udq.py
+++ b/udq_actionx/udq.py
@@ -5,7 +5,7 @@
 #   DEFINE WULPRU (WLPR OPU01 - 300) * 0.80 /
 # /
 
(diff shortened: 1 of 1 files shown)


blattms:
Is there any dependency between the two PRs (like one does not compile without the other?

lisajulia:
Thanks for notifying! I'm closing this now since we have decided in the last meeting to actually keep the actionx_callback

blattms:
Seems like the script is used in udq_actionx/PYACTION_WCONPROD which we run a regression one. Maybe we should even have two versions (one with and one without callback) and ideally even compare their results.

PR #1141, opened 2024-03-25
Test lgr with deformed cells

aritorto:
This PR adds two tests:
- refine a single deformed cell,
- refine 2 deformed cells. 

Currently, LGRs are supported only for CpGrid, with the set of cells to be refined being regular (block of cells with certain uniform widths, lengths, and heights). However, the refinement of a single deformed cell can be supported as well. The only modification needed is to refactor the requirements of the LGRs. Namely, when the size of the LGR is only 1 (one cell), the only restriction for such a cell is to have 8 different corners.  

When the set of cells to be refined contains more than one deformed cell, more work is required.


aritorto:
jenkins build this please

aritorto:
Not meaningful for real applications. Only for internal test.

PR #1142, opened 2024-04-05
Added models for regression tests of MULTXYZ

blattms:
This is meant to be used together with OPM/opm-common#3997


bska (approved):

PR #1145, opened 2024-04-08
Simplify pyaction/pyaction_wconprod_insert_kw.py script

lisajulia:
Simplified script, still produces the same output

### DIFF
--- a/pyaction/pyaction_wconprod_insert_kw.py
+++ b/pyaction/pyaction_wconprod_insert_kw.py
@@ -16,7 +16,7 @@ if (not 'setup_done' in locals()):
 
 current_time = schedule.start + datetime.timedelta(seconds=summary_state.elapsed())
 
-days_since_last_update = [(current_time - last_updated_at[0]).total_seconds()/3600/24, (current_time - last_updated_at[1]).total_seconds()/3600/24, (current_time - last_updated_at[2]).total_seconds()/3600/24]
+days_since_last_update = [(current_time - i).total_seconds()/3600/24 for i in last_updated_at]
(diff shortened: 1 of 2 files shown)


lisajulia:
@blattms: Can you have another look here? Thanks!

blattms (commented):
So basically globals() and locals() are the same here as we are at module level?

pyaction/pyaction_wconprod_insert_kw.py
```diff
@@ -1,84 +1,55 @@
...
+    all_wells_shut = False
+    setup_done = True
 
-    current_time = schedule.start + datetime.timedelta(seconds=summary_state.elapsed())
+current_time = schedule.start + datetime.timedelta(seconds=summary_state.elapsed())
 
-    act01_days_since_last_execution = (current_time - globals()["act01_last_executed_at"]).total_seconds()/3600/24  
-    act02_days_since_last_execution = (current_time - globals()["act02_last_executed_at"]).total_seconds()/3600/24  
-    act03_days_since_last_execution = (current_time - globals()["act03_last_executed_at"]).total_seconds()/3600/24  
+days_since_last_update = [(current_time - last_updated_at[0]).total_seconds()/3600/24, (current_time - last_updated_at[1]).total_seconds()/3600/24, (current_time - last_updated_at[2]).total_seconds()/3600/24]
```
blattms, line 19: Newbee question :sweat_smile:: Would this also work?
```suggestion
days_since_last_update = [(current_time - i).total_seconds()/3600/24 for i in last_updated_at]
```

lisajulia:
> So basically globals() and locals() are the same here as we are at module level?

Yes, here they are, yet inside the run function, I had to use globals, because otherwise the variables would have been only at local scope.

lisajulia:
jenkins build this please

PR #1148, opened 2024-04-12
[fix] Added missing include files for MULTXYZ output tests.

blattms:
Fixes fallout from OPM/opm-simulators#5274 (and #1142).


bska:
The `MULTXYZ*` cases [also have](https://github.com/OPM/opm-tests/blob/8a1fa495d11c3873f9049b40f9d626afebf393fb/mult/MULTXYZ05.DATA#L77-L78)
```
INCLUDE
 './include/t1_3x2x3.grdecl' /
```
but I don't see the `t1_3x2x3.grdecl` file here, or in the `include` directory in [current master](https://github.com/OPM/opm-tests/tree/8a1fa495d11c3873f9049b40f9d626afebf393fb/mult/include).  Is that file supposed to be somewhere else?

blattms:
Good catch. Added now,

bska (approved):

PR #1151, opened 2024-04-17
Use opm_embedded.OpmLog.info instead of print in the python functions

lisajulia:
With this, you can pipe the simulation output and the python printouts are the correct spot


bska:
jenkins build this please

bska (approved):
Greatly appreciated.  I'll merge into master.

PR #1155, opened 2024-04-20
Test case for WECON item10 = RATE

tskille:
The minimum rate limit should not be applied for wells under group control if the target production rate is 0.0.

The current version of OPM Flow does not honour this exception.


bska (approved):
Looks good to me.  I'll merge into master.

PR #1171, opened 2024-05-30
Test to verify valid SATNUM in AQUNUM fixes any invalid SATNUM value

vkip:
Specifically, checks that cells with a valid SATNUM specified in AQUNUM pass internal checks even with invalid values in input SATNUM array in the REGIONS section.

Ref https://github.com/OPM/opm-common/pull/4082


bska (approved):

PR #1172, opened 2024-05-31
two new test cases with keyword WTMULT

tskille:
Two more test cases using keyword WTMULT. Case WTMULT-03 is using keyword WTMULT inside an actionx block.


bska (approved):

PR #1174, opened 2024-06-04
Test case with UDQ for group quantities

tskille:
This test case includes user-defined group quantities (UDQ). This is not fully supported yet but hopefully this will come soon.


bska (approved):
Thanks a lot&ndash;this will help expedite adding the feature.  I'll merge into master.

PR #1175, opened 2024-06-05
Test case for supporting instantaneous rates in extended network 

vkip:
To test https://github.com/OPM/opm-common/pull/4094 + https://github.com/OPM/opm-simulators/pull/5410


GitPaean (approved):
The test case looks fine to me. 

Maybe in the file name we can reflect that we are setting the item 3 to be 'NO' for WEFAC and GEFAC in case we have a test case later that using the default values ('YES') for these two keywords.

GitPaean:
Thanks for the update. It looks good to me now.  I am getting it in.

PR #1191, opened 2024-07-02
changed tolerance in NETBALAN

plgbrts:


GitPaean:
The change is so that the upstream development can solve this case.  Thanks for the update.

PR #1199, opened 2024-07-25
Add test for ACTIONX with COMPDAT

lisajulia:


blattms:
@gdfldm, would take a look at these tests to check whether the combination of COMDAT and COMPLUMP makes sense, please

gdfldm:
> @gdfldm, would take a look at these tests to check whether the combination of COMDAT and COMPLUMP makes sense, please

In ACTIONX_COMPDAT_SHORT, COMPLUMP is used for wells PROD2 and INJ1.

PROD2: COMPLUMP is used to lump the well connections to create upper and lower completions at the start of the run in ACTIONX ACT-01. The lower completion is opened by ACTIONX ACT-02 on 1/12/2018. The upper completion doesn't seem to be used. This seems like a reasonable test although clearly in practice you don't need actions to do this.

INJ1: again COMPLUMP is used to lump the well connections to create upper and lower completions. But the completions don't seem to be used anywhere? (previously they were used by ACT-04 in ACTIONX_COMPDAT, which has been deleted in this case). So this instance of COMPLUMP could probably be removed.

lisajulia:
> > @gdfldm, would take a look at these tests to check whether the combination of COMDAT and COMPLUMP makes sense, please
> 
> In ACTIONX_COMPDAT_SHORT, COMPLUMP is used for wells PROD2 and INJ1.
> 
> PROD2: COMPLUMP is used to lump the well connections to create upper and lower completions at the start of the run in ACTIONX ACT-01. The lower completion is opened by ACTIONX ACT-02 on 1/12/2018. The upper completion doesn't seem to be used. This seems like a reasonable test although clearly in practice you don't need actions to do this.
> 
> INJ1: again COMPLUMP is used to lump the well connections to create upper and lower completions. But the completions don't seem to be used anywhere? (previously they were used by ACT-04 in ACTIONX_COMPDAT, which has been deleted in this case). So this instance of COMPLUMP could probably be removed.

Thanks @gdfldm, I've updated the test accordingly

PR #1203, opened 2024-08-06
Keywords (OPTIONS and PIMULTAB) not supported were commented.

Vomartins:
Hello everyone,

The unsupported keyword OPTIONS and the keyword PIMULTAB were being used in Norne cases. I commented these keywords and the PIMULTAB related keyword PIMTDIMS.

Here is the error message screenshot that I got before comment there keywords.
1A_STDW

Flow version 2024.10-pre


alfbr:
jenkins build this please

alfbr:
Thanks for the contribution! These keywords were ignored earlier, but now they will cause the simulation to stop. Hence, there should be no changes to the simulation results due to this change. Merging once jenkins is green.

PR #1207, opened 2024-08-14
First GSATPROD reference model

tklov11:
Added a box model demonstrating the use of a new keyword GSATPROD for inclusion of satellite production rates in the simulator.


alfbr:
Indeed, but I do miss a sentence or two below the copyright notice in the DATA-file to explain what is tested.

PR #1208, opened 2024-08-16
Test case for well bottom hole pressure and average pressure

tskille:
using a representative well bore density for datum depth corrections.

The current version of flow is not doing this depth correction correct. However, @bska is working on a fix for this. 

https://github.com/OPM/opm-simulators/pull/5479


bska (approved):
Very good.  I'll merge into master.  Thanks a lot for the contribution.

PR #1225, opened 2024-09-06
Two additional test cases with keyword pinch and item 4 = ALL

tskille:
These are identical to T1B3 except that this includes MULTZ in schedule section.


bska (approved):
Thanks a lot for expanding our suite of test cases for this feature.  This looks good to me and I'll merge into master.

PR #1227, opened 2024-09-12
LGR Test cases (cornerpoint with non active cells)

Negar-Khoshnevis:
I added the three test cases for LGR Test cases (cornerpoint with non active cells)


alfbr:
Thanks for the contribution, this is highly appreciated! I have not yet tested the LGR part of this, but we can do that when we enable these as regression tests for Flow. Everything looks perfect at this point in time, so I will merge this. However, please double check the copyright notice on the first model. It reads TNO, but should maybe be Equinor?

Negar-Khoshnevis:
The copyright is changed and updated.

alfbr:
Thanks for the clarification, merging!

PR #1240, opened 2024-10-09
Test models using well template names as input.

tskille:
Testing ACTIONX and WCONPROD with templated well names and well lists.


bska (approved):
Thanks a lot for putting together this set of example models&ndash;it is greatly appreciated.  I'll merge into master.

PR #1245, opened 2024-10-28
commenting out COMPS NCOMPS CNAMES

GitPaean:
for testing how the compositional keywords are used in blackoil simulation

### DIFF
--- a/co2store/CO2STORE_GASWAT.DATA
+++ b/co2store/CO2STORE_GASWAT.DATA
@@ -19,10 +19,6 @@ METRIC
 CO2STORE
 
 
-COMPS
-2
-/
-
 DIMENS
    50 50 10  /
 
@@ -81,18 +77,6 @@ ROCK
 
 RTEMP
   80 
-/
-
-
-NCOMPS
-
-       2
-/
-CNAMES
-
-   'CO2'
-   'H2O'
-
 /
 
 ZMFVD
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

GitPaean:
It looks like those keywords can be removed.  Will do the same with H2STORE_GASWAT.

GitPaean:
@totto82 , do you know those keywords are used in the simulation, can they be removed?

GitPaean:
jenkins build this please

totto82:
They are needed if DENAQA or VISCAQA are used to map the coefficients in these keywords to the right components. 

Since CO2STORE and H2STORE is a special case where compositional and blackoil are mixed maybe we need a check if(!co2store) output compositional stuff. Or even better would be to compute the needed output for CO2STORE to make it compatible with compositional type output. As a first attempt I would suggest a if(!co2store) type solution.

GitPaean:
> They are needed if DENAQA or VISCAQA are used to map the coefficients in these keywords to the right components.
> 
> Since CO2STORE and H2STORE is a special case where compositional and blackoil are mixed maybe we need a check if(!co2store) output compositional stuff. Or even better would be to compute the needed output for CO2STORE to make it compatible with compositional type output. As a first attempt I would suggest a if(!co2store) type solution.

Thanks for the prompt response. I am not aware of the setup of keywords `DENAQA or VISCAQA `.   Let me assume both `co2store` and `h2store` are pure black oil implementation for now.  we might need to revisit these at a later stage when we get there.

PR #1248, opened 2024-11-18
Reduce stepsize in 9_3D_GINJ_GAS_MAX_EXPORT_MSW for after May, 20th 2019

lisajulia:
to better compare the test compareECLFiles_flow+9_3D_GINJ_GAS_MAX_EXPORT_MSW for https://github.com/OPM/opm-simulators/pull/5741

### DIFF
--- a/model2/9_3D_GINJ_GAS_MAX_EXPORT_MSW.DATA
+++ b/model2/9_3D_GINJ_GAS_MAX_EXPORT_MSW.DATA
@@ -487,6 +487,9 @@ DATES   -- 11
  20 'MAY' 2019 /
 /
 
+NEXTSTEP
+2    NO / 
+
 -- loosing injector INJ1, only remaining injector is INJ2
 WELOPEN
  'INJ1' 'SHUT' /


alfbr:
This change is fine I believe. However, do we actually honour the NEXTSTEP keyword? Since we ignore TUNING by default, it does not really make sense for me to honour NEXTSTEP given that the numerics of Flow is very different from other simulators. In any case, I believe you can achieve the same by an additional DATES, i.e., instead of NEXTSTEP you could do:

DATES
 22 'MAY' 2019 /
/

lisajulia:
Thanks! Yes, I think you're right!
Yet also my approach of changing the test file is wrong: @GitPaean told me to compare the failing test with --solver-max-time-step-in-days=1 locally and then if that is ok, we update the reference file - so I'll do that.

alfbr:
Either way works for me, but I actually think your option is better than doing custom command line options in the cmake files (keeping them clean is a plus, while exposing the time step constraint in the test itself is also a plus). In addition, reducing max time step to one day will probably make the test much slower to run, while your actual need is probably to limit the time step around the time when the control kicks in. In any case, your call, just sharing my five cents.

PR #1249, opened 2024-11-18
Adjust the timestepsizes in 3 tests

lisajulia:


lisajulia:
Closing for the same reason as #1248

PR #1252, opened 2024-11-21
added equals+minpvv to existing test model

tskille:
Equals + MINPVV should remove 3 additional cells from the model Added one time step to the model in order to make the model run.


tskille:
This is currently not supported in OPM Flow. The existing test model is not used for regression testing today.

bska (approved):
Thanks a lot for the contribution.  This is, indeed, a use case that's not currently supported in the simulator.  I'll merge into master.

PR #1272, opened 2025-01-02
Add two-phase gas/water case with thermal.

atgeirr:
This adds thermal effects to the existing SPE1CASE2_GASWATER.DATA. The primary intended usage is for development of a proof-of-concept for props and assembly on GPU, where our first target is SPE11-like gas/water/thermal simulation. Tests using this file may eventually be added.


bska (approved):
Thanks, I'll merge into master.  For future reference, though, we often prefer to add a descriptive comment highlighting the feature under test&ndash;see e.g., [`ACTIONX_WELL_TEMPL-05`](https://github.com/OPM/opm-tests/blob/96da209393eb6126bf38dd0a946f476e4bc316d4/actionx/ACTIONX_WELL_TEMPL-05.DATA#L8-L25)&ndash;to assist the reader in their understanding of the setup.

PR #1279, opened 2025-01-10
Test case with a UDQ holding maximum string length (input)

tskille:
There is a maximum length = 128 (16 x 8). This is related to the restart format.

This test case includes the UDQ **FU_TEST** which holds exactly 128 characters. OPM flow is currently not supporting this deck. The simulation stops and the following Error message is reported.

```
ERROR: Uncaught std::exception when running tasklet: DEFINE expression for UDQ FU_TEST is too long.
  Number of characters 164 exceeds upper limit of 128.
  Expression: (ROIP_RE2 '2') / (FURE2) * (0.3 * (RGPR_RE2 '1') + (RGPR_RE2 '2') + (RGPR_RE2 '3')) / (GEFF 'TEST') * (1 + FUGASX) + ((1 * TIME * TIME - 75 * TIME + 1250) * 50.123).
```

Notice that if you remove keyword RPTRST in the Solution section, the simulation will run and simulation results are good.

### DIFF
--- a/udq_actionx/UDQ_REG-02.DATA
+++ b/udq_actionx/UDQ_REG-02.DATA
@@ -1,0 +1,334 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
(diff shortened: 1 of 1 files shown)


bska:
Thank you for providing this test case. I will start working on this problem right away.

PR #1286, opened 2025-01-17
Add test case with GSATPROD and GCONPROD

totto82:
This is GSATPROD1.DATA with the addition of GCONPROD


totto82:
jenkins build this please

PR #1294, opened 2025-02-18
add RESIDUAL to test output

akva2:
It was discovered that there are no regression tests with RESIDUAL output. This remedies that.


bska (approved):
Thanks.  I'll merge this now to prepare for merging the upstream PR.

PR #1296, opened 2025-02-19
add BFLOWIJK to test output

daavid00:
It was discovered that there are no regression tests with BFLOW Summary output. This remedies that.


akva2:
merging to prepare for merging the upstream.

PR #1299, opened 2025-02-26
Test case with UDAs

tskille:
Including UDAs for keywords GCONPROD, GCONINJE, WCONPROD and WCONINJE.


tskille:
UDA functionality is already supported in simulator, but eclipse compatible restart is not yet supported. Hopefully this can be implemented soon.

bska (approved):
Thanks a lot.  This will aid in developing restart file support.

PR #1310, opened 2025-03-24
Add two tests with multisegment wells that are distributed across two processes and have shut perforations

lisajulia:
For this, run the case with flow_distribute_z on 4 processes, the well PROD1 has 12 perforations, 5 on rank 0, 7 on rank 1.


GitPaean:
Can you explain what the `BORDER` means in the filename?

lisajulia:
> Can you explain what the `BORDER` means in the filename?

It means that the open perforation is at the border between two processes.

PR #1316, opened 2025-03-31
Add test case for network + wtest

totto82:
See https://github.com/OPM/opm-simulators/pull/6116


GitPaean:
Please do not forget to add the regression test in opm-simulators to utilize this DATA file.

PR #1327, opened 2025-04-03
remove KRW from the 1D_COMP.DATA

GitPaean:
the case does not have water, there is no point to have KRW there.


GitPaean:
The impact of this change is rather limited. I am merging it in now.

PR #1335, opened 2025-04-29
Add Small Fracturing Example

bska:
This is in preparation of introducing geo-mechanical analysis and fracturing support into the mainline Flow simulator.

The initial example is a small, 15-by-15-by-14 Cartesian grid with uniform properties and a single injector in the centre of the formation injecting low-temperature water.


bska:
jenkins build this please

bska:
I'm closing this now. The geomechanical solves with fracturing have progressed beyond what this example illustrates so it's better to create a new example than to give false impressions.  I will do so when we're closer to integrating the real simulation support.

PR #1344, opened 2025-06-02
adding tempeature related WELL SUMMARY keyword for SPE1CASE2_THERMAL case

GitPaean:
WTPRHEA, WTPTHEA, WTIRHEA and WTITHEA are not implemented yet, so it outputs zero for them for now.  Please let me know whether we can add them or we should wait for this type of situation. 

WTPCHEA and WTICHEA is outputting reasonable values up to verification.


GitPaean:
jenkins build this please

akva2:
actually it's already branched. so it's all good.

PR #1351, opened 2025-06-18
Updating TUNING values due to bugfix in opm-simulators #6332

daavid00:
The `TUNING` keyword was assigned for min time step `TSFMIN` instead of `TSMINZ`, which default values are significantly different, then  by fixing the bug this two test fail, therefore setting the values `TSMINZ` in the deck  (see https://github.com/OPM/opm-simulators/pull/6332 for further details)


bska:
Please give a better description.  The current (see this other place) is not good enough.

bska (commented):
I'm not opposed to this, but why do we need minimum time step sizes that are this small? That suggests that there's something quite unhealthy about these cases.

daavid00:
> I'm not opposed to this, but why do we need minimum time step sizes that are this small? That suggests that there's something quite unhealthy about these cases.

You are right, both runs struggle, e.g., for 02:-WGRUPCON.DATA:

```
Number of newton iterations reached its maximum try to continue with relaxed tolerances:  MB: 1.0e-06 CNV: 1.0e+00
  20  1.489e-07  2.458e-07  1.236e-07  1.601e-03  2.234e-04  2.522e-04
Caught Exception: [/Users/dmar/Github/opm/opm-simulators/opm/simulators/flow/NonlinearSolver.hpp:176] Solver convergence failure - Failed to complete a time step within 20 iterations.
Solver convergence failure - Iteration limit reached
Timestep chopped to 0.015314062500000008 days


Starting time step 9, stepsize 0.0153141 days, at day 36.6675/40, date = 06-Feb-2020
   Well C-1H did not converge in 50 inner iterations (12 switches, 12 status changes).
estimateOperableBhp: Did not find operable BHP for well C-1H
 well C-1H gets STOPPED during iteration
```

PR #1368, opened 2025-08-18
Two test cases with keyword GSATINJE

tskille:
These two test cases both include keyword GSATINJE. This keyword is currently not supported in OPM Common.  Hopefully this will be implementet soon.


tskille:
The test case GSATINJ_GAS-01.DATA has been updated and this now runs fine.

PR #1388, opened 2025-09-02
Avoid using def run (the old way..)

totto82:
See discussion in https://github.com/OPM/opm-simulators/pull/6435


bska (approved):
Excellent!  I'll merge into master.

PR #1397, opened 2025-09-15
Switch to DATES Based Time Stepping

bska:
This is in preparation of testing simulation restart for GSATPROD.

The "rst_deck" utility doesn't really handle TSTEP-based time stepping when there are multiple steps in a single TSTEP keyword.


bska:
jenkins build this please

PR #1413, opened 2025-10-14
Set calc to 1 in model3 to avoid internal error

daavid00:
In master, [3A_WSEGSICD_MODEL3.DATA](https://github.com/OPM/opm-tests/blob/master/model3/3A_WSEGSICD_MODEL3.DATA), [3C_WSEGAICD_MODEL3.DATA](https://github.com/OPM/opm-tests/blob/master/model3/3C_WSEGAICD_MODEL3.DATA), and [4_NUMAQ_MODEL3.DATA](https://github.com/OPM/opm-tests/blob/master/model3/4_NUMAQ_MODEL3.DATA) ends with:

> ```
> In /Users/dmar/Github/opm/todostests/model3/include/wells.sch3 line 73
> Internal error: Zero connection length is found. No way to update scaling factor for this SICD segment
> Problem with keyword WSEGSICD
> ```

This PR sets item 11 in `WSEGSICD` and `WSEGAICD` to 1, i.e., using the absolute value of the `ICDLEN` (item 5).


GitPaean:
> Hi @GitPaean, thanks for looking at this. Should I update the PR changing COMPSEGS before WSEGAICD/WSEGSICD then?

It is possible while I am more thinking maybe we should decide the behavoir we/users want first.  If it is confirmed that we should be able to run the cases, then we need to update the code instead of modifying the cases.

daavid00:
Closing this PR since we should support the keyword setup with WSEGAICD first then COMPSEGS later.

PR #1426, opened 2025-11-12
Add SPE1CASE2 TEMP case

totto82:
Adds a test case for the TEMP option. SEE https://github.com/OPM/opm-simulators/pull/5854


totto82:
jenkins build this please

atgeirr:
Test failure expected since the THERMAL case has been shortened. Doing a data update should be appropriate.

atgeirr (approved):

bska:
>  update_data

Alas, that doesn't work when triggered from opm-tests.  If you need to update reference solutions, then the `update_data` request must be issued from an upstream repository.

totto82:
I have triggered a update_data from https://github.com/OPM/opm-common/pull/4825

totto82:
The PR is approved, and Jenkins is happy. I will merge.

PR #1440, opened 2025-12-04
Test cases for TPSA geomechanics

svenn-t:
Simple 5x5x5 test cases for TPSA using lagged and fixed-stress coupling schemes to Flow.

See https://github.com/OPM/opm-simulators/pull/6650 and https://github.com/OPM/opm-common/pull/4846.

PR in draft mode, since both companion PRs are in draft mode.


totto82:
Thanks for the case. With enable-opm-rst-file=true it gives output.

PR #1502, opened 2026-02-24
Test for overlaping NNC and EDITNNC

daavid00:
Adding these decks to the regression tests, to cover overlapping NNC with EDITNNC and MULTREGT.

Depends on https://github.com/OPM/opm-common/pull/5025


gdfldm (approved):
Looks good to me.

PR #1514, opened 2026-03-27
Correcting name (overlapping instead of overlaping)

daavid00:
Correcting name (overlapping instead of overlaping). These tests are added to the CI in https://github.com/OPM/opm-simulators/pull/6968


blattms:
Seems clean and not used yet.
I am merging this. Not sure whether it will break #1515, though. Maybe we should redo that one.

PR #1520, opened 2026-04-30
removing OPTIONS so that flow can run directly for POLYMER-01.DATA and POLYMER-02.DATA

GitPaean:
not sure what the policy is for this. 

But I think the DATA files in opm-tests should be able to run with flow with minimum arguments.


bska:
> I think the DATA files in opm-tests should be able to run with flow with minimum arguments.

I agree. Just for clarity, however: Can you run these with `--parsing-strictness=low`?

GitPaean:
> > I think the DATA files in opm-tests should be able to run with flow with minimum arguments.
> 
> I agree. Just for clarity, however: Can you run these with `--parsing-strictness=low`?

yes, they can be run with `--parsing-strictness=low`.

bska (approved):
>> Can you run these with `--parsing-strictness=low`?
>
> yes, they can be run with `--parsing-strictness=low`.

Okay, in that case there's absolutely no reason to have these `OPTIONS` settings which we'll never support.

I'll merge into master.

PR #1522, opened 2026-05-18
Test case for injection network

tskille:
One test case for gas injection and one for water injection.  Injection network (keyword GNETINJE) is currently not supported in OPM Flow.


bska (approved):
Thanks a lot for providing these example cases. These will be very useful for developing support for `GNETINJE` and the associate summary vectors and I'll merge into master.

PR #1523, opened 2026-05-19
Test case for GECON and maximum water-gas ratio

tskille:
Simulator should shut worst offending well when maximum water-gas ratio for group field is reached. This case is currently not supported in OPM FLow.


bska (approved):
Thanks a lot for providing an example case. I'll merge this into the master branch.

PR #1525, opened 2026-05-21
Test case for EQLOPTS + IRREVERS

tskille:
EQLOPTS + IRREVERS is not yet supported in OPM Flow.


bska (approved):
Thanks a lot for providing a test case.  I'll merge into master.

PR #1526, opened 2026-05-26
3 test cases for keyword CECON

tskille:
Keyword CECON is currently not supported in OPM Flow.


bska (approved):
Thank you very much for providing these example models.

I'll merge into master.

PR #1546, opened 2026-07-10
adding the test WECON_PLUSCON_COMPLUMP.DATA

GitPaean:
the base is from model6, but is for testing WECON +CON workover combined with COMPLUMP. 

It is one of the development model for PR https://github.com/OPM/opm-simulators/pull/7141 .


GitPaean:
this is good test case for WECON +CON development in https://github.com/OPM/opm-simulators/pull/7141, but it is based on model 6. I struggled some to where and how to add it. Suggestions will be welcomed.

bska (approved):
> it is based on model 6. I struggled some to where and how to add it

I think it's fine to put it here (`opm-tests/wecon_wtest`) since it's for exercising `WECON`.  I'll merge into master.

PR #1549, opened 2026-07-28
Add missing record terminator to AQUFET in SPE1CASE2_AQUFET.DATA

### LINKED ISSUE #1548: spe1/SPE1CASE2_AQUFET.DATA does not load: missing AQUFET terminator, and AQUFET is unsupported by flow
`spe1/SPE1CASE2_AQUFET.DATA` fails to load in OPM Flow. There are two independent defects, and the deck appears to be unreferenced by any test list — which is presumably why neither has surfaced.

Found while using opm-tests as a false-positive corpus for a VS Code deck-linting extension: this deck was the one case where the linter's warning turned out to be correct.

## 1. `AQUFET` is missing its terminating `/`

```
285| AQUFET
286| 	8400 4800 1200E6 3E-6 10E6 1 1 10 1 1 3 3 'J-' /
287| SUMMARY
```

`AQUFET`'s keyword definition declares `items` with no `size`, so `ParserKeyword::initSize` classifies it `SLASH_TERMINATED`. The parser only lets a recognised keyword end the current one when `RawKeyword::can_complete()` is true, and that requires `UNKNOWN` size type or `m_min_size < m_fixedSize` — for `SLASH_TERMINATED` both are 0, so it returns false.

`SUMMARY` is therefore consumed as record data rather than starting a new keyword, and the keyword runs on to the next lone `/` at line 298, swallowing `SUMMARY`, `FOPR`, `WGOR` and `'PROD'`:

```
Reading AQUFET   in SPE1CASE2_AQUFET.DATA line 285

Error: Problem with keyword AQUFET
In SPE1CASE2_AQUFET.DATA line 285.
Failed to parse the record 2 out of 4 of the keyword AQUFET
Malformed floating point number 'SUMMARY'
```

Note *"record 2 out of 4"* — four records were collected from a block meant to hold one.

Adding a standalone `/` after line 286 clears this error.

## 2. `AQUFET` is not supported by flow

With the terminator added, the deck gets past parsing and then hits:

```
Error: Unsupported keywords or keyword items:

  AQUFET: keyword not supported
  In file: SPE1CASE2_AQUFET.DATA, line 285
  Use the AQUFETP keyword instead
```

So the deck cannot run as written regardless of the syntax fix. The parser recognises `AQUFET`; the simulator does not implement it.

## 3. The deck is unreferenced

`SPE1CASE2_AQUFET.DATA` does not appear in any test list, manifest, CMake file, or script in the repository — the only match outside the file itself is `.git/index`.

## Suggested resolution

Since the deck exercises a keyword flow does not support, and nothing runs it, deleting it may be cleaner than repairing it. The alternative is converting it to `AQUFETP`, but that is a semantic rewrite rather than a syntax fix: `AQUFET` carries its connection box inline (`LOWER_I`…`FACE_INDX`), whereas `AQUFETP` takes an `AQUIFER_ID` with connections supplied separately via `AQUANCON`.

Happy to open a PR for whichever direction you prefer.

## Reproduction

```
flow SPE1CASE2_AQUFET.DATA --enable-dry-run=true
```

`flow 2026.10-pre`, opm-tests at `c48d6ecf`.

magnesj:
AQUFET is a slash-terminated keyword, so its block needs a standalone `/` to end it. Without it the parser consumed `SUMMARY`, `FOPR`, `WGOR` and `'PROD'` as AQUFET record data and failed with:

```
Error: Problem with keyword AQUFET
In SPE1CASE2_AQUFET.DATA line 285.
Failed to parse the record 2 out of 4 of the keyword AQUFET
Malformed floating point number 'SUMMARY'
```

This adds the missing terminator so the deck parses, as requested in #1548. Note that flow still rejects the deck at a later stage since AQUFET itself is unsupported ("Use the AQUFETP keyword instead") — that part of the issue is left as is per the discussion there.

Fixes #1548


alfbr:
Huge thanks, this is going in.

bska:
> AQUFET is a slash-terminated keyword, so its block needs a standalone `/` to end it.

Very strictly speaking it is not.  OPM's parser, however, is unable to handle analytic aquifer keywords that don't have an otherwise spurious terminator&ndash;i.e., the same issue holds for `AQUCT` and `AQUFETP`.  Adding this terminator is therefore a workaround for a deficiency in the OPM parser.  Other parsers don't have this problem.

PR #1550, opened 2026-08-04
Test case with keyword WECON + item 14 (minimum liquid rate)

tskille:
WECON + Item 14 is currently not supported in OPM Flow.


GitPaean:
Thanks for the update. It looks good and it runs fine. I am getting it in now.

PR #1562, opened 2026-08-16
Two test cases with keyword GSATPTAB

tskille:
These models are using keyword GSATPTAB which is currently not supported in OPM Flow.uses


bska (approved):
Thank you for the example models. We will start scoping out the work needed for `GSATPTAB` as soon as possible.

I'll merge this into the master branch.

PR #1563, opened 2026-08-17
Test case with WCONPROD+STOP and WELOPEN+STOP

tskille:
This seems to be not covered in the current version of opm-tests and OPM Flow is not supporting this correctly.


bska (approved):
Thanks.  I'll merge this into the master branch.

PR #1573, opened 2026-08-24
specifying explicitly the item 2 of WELDRAW, we do not allow default item 2 in WELDRAW

GitPaean:
Item 2 has no default, so leaving it out is rejected rather than being a way of lifting the limit.  Raise D-1H well above its natural drawdown instead,  which lifts the limit in practice and leaves the results unchanged.

### DIFF
--- a/weldraw/WELDRAW-01-MSW.DATA
+++ b/weldraw/WELDRAW-01-MSW.DATA
@@ -32,8 +32,10 @@
 --   1 Dec 2018 - 1 Mar 2019    WELDRAW active: 5.0 bar on D-1H, 10.0 bar on
 --                              D-2H.  Both limits are below the natural
 --                              drawdown, so both wells are cut back.
---   1 Mar 2019 - 1 Jun 2019    the limit on D-1H is removed by defaulting
---                              item 2, and D-2H is tightened to 6.0 bar.
+--   1 Mar 2019 - 1 Jun 2019    the limit on D-1H is raised to 100.0 bar,
+--                              well above its natural drawdown, so that it
+--                              no longer binds, and D-2H is tightened to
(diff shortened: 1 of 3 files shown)


GitPaean:
It is the new finding with some new cases. I make it as draft because it only goes in if the design is confirmed on the upstream PR https://github.com/OPM/opm-common/pull/5305 .  Otherwise, the comments should be revised following the designed behavoir, potentially, we might throw with default item 2 of WELDRAW, then the DATA file needs to adapt to make it runable.

GitPaean:
Now the case [WELDRAW-01.DATA‎](https://github.com/OPM/opm-tests/pull/1573/changes#diff-8f555a6a8732382ea150065b5007b31f7118d654b1d316ffa71c3dcc4fcd88d4) is modified so that it reflects the behavoir implemented in upstream PR, https://github.com/OPM/opm-common/pull/5305 .

GitPaean:
now we agreed with the behavoir, the DATA files are amended to reflect the agreed behavoir.

bska:
jenkins build this please

GitPaean:
just some information, these cases are not registered as regression tests in the master branch yet. The regression tests are added in https://github.com/OPM/opm-simulators/pull/7346 , which are under reviewing.

GitPaean:
https://ci.opm-project.org/job/opm-tests-PR-builder/289/

image


not sure what this is about, but putting here for information before I rebase the branch.

bska:
> not sure what this is about,

It's **probably** a race with #1566/OPM/opm-simulators#7331.

bska:
jenkins build this please

bska (approved):

PR #1574, opened 2026-08-24
two compositional cases with EQUIL + ZMFVD

GitPaean:
they are the development cases for the development in the https://github.com/OPM/opm-common/pull/5283 and https://github.com/OPM/opm-simulators/pull/7306 .


bska (approved):
Time spans are rather short, but as a stepping stone these look good to me.  I'll merge into master.

GitPaean:
> Time spans are rather short, but as a stepping stone these look good to me. I'll merge into master.

yes. We still have very limited cases, they are among the ones that we have reference results.  It is for equilibration development.

PR #1576, opened 2026-08-25
putting compositional equilibration cases in a equilibration folder under compositional

GitPaean:
for better organization since more cases will come.


bska (approved):

PR #1579, opened 2026-09-04
Sloping aquifer test model

totto82:
See https://github.com/OPM/opm-simulators/pull/7327


bska (approved):

PR #1581, opened 2026-09-04
Add test for THPRESFT keyword

gdfldm:
This case is based on FAULTS_MODEL_1.DATA. Faults are added between the equilibration regions, and THPRES is replaced with the THPRESFT keyword. The threshold pressures for the faults are based on the values reported in FAULTS_MODEL_1.PRT.


bska (approved):
Very good.  I'll merge into msater.

PR #1585, opened 2026-09-07
Add tests for WGRUPCON default rate and zero guiderate potential fallback

vkip:


model2/include/PVT-DRY-GAS-LIVE-OIL.INC
```diff
@@ -0,0 +1,338 @@
+-- This reservoir simulation deck is made available under the Open Database
+-- License: http://opendatacommons.org/licenses/odbl/1.0/. Any rights in
+-- individual contents of the database are licensed under the Database Contents
+-- License: http://opendatacommons.org/licenses/dbcl/1.0/
+
+-- Copyright (C) 2018 Equinor
+
```
totto82, line 7: 2018? Is this just a copy of some other PVT file?

totto82 (approved):
Thanks for the tests.

model2/include/PVT-DRY-GAS-LIVE-OIL.INC
vkip: Good point, I'll update. (Not just a copy, but a modified version of another file).

PR #1593, opened 2026-09-09
Add regression coverage for compositional PSAT output

GitPaean:
Adds `PSAT` to the `RPTRST` configuration of the registered `1D_COMP` and `SIMPLE_COMP_SSHIFT` compositional cases, including the `SOLUTION` section so the initial restart record is covered.

No registered compositional regression previously requested this field. The existing equilibration decks that request it are not registered, while black-oil cases containing `PSAT` do not exercise `OutputCompositionalModule`.

The two cases cover complementary states:

- `1D_COMP` starts with two hydrocarbon phases, where `PSAT` equals the cell pressure. Intermediate single-phase states exercise the saturation-pressure solver, while late near-pure CO2 states at 150 °C have no saturation boundary and exercise the zero-valued unresolved-result path.
- `SIMPLE_COMP_SSHIFT` covers the same output with volume shifts enabled. Its column passes through two-phase, liquid-only, and vapour-only states, exercising the cell-pressure, bubble-pressure, and dew-pressure paths.

Both decks use `RPTRST` in the `SOLUTION` section because `RPTSOL` is normalised against the `RPTSCHED` mnemonics, which do not include `PSAT`.

Reference data is intentionally omitted and is left to the Jenkins `update_data` workflow, which regenerates it using a build containing the compositional `PSAT` implementation.

Requires [OPM/opm-simulators#7389](https://github.com/OPM/opm-simulators/pull/7389), which uses the saturation-pressure solver merged in [OPM/opm-common#5283](https://github.com/OPM/opm-common/pull/5283).


GitPaean:
jenkins build this failure_report please

https://ci.opm-project.org/job/opm-tests-PR-builder/294/

bska:
> References are regenerated for `flow_comp` and `flowexp_comp`.

Did you create those on your own machine?  If so, there's a chance that they won't apply to the CI system.  I've experienced on a number of occasions that my machine computes result sets that are ever so slightly different from those of the CI system's.  Different CPUs, different compilers (or compiler versions), different build flags &c could all be contributing to changes that make locally generated solutions not match those of the CI system.

GitPaean:
> > References are regenerated for `flow_comp` and `flowexp_comp`.
> 
> Did you create those on your own machine? If so, there's a chance that they won't apply to the CI system. I've experienced on a number of occasions that my machine computes result sets that are ever so slightly different from those of the CI system's. Different CPUs, different compilers (or compiler versions), different build flags &c could all be contributing to changes that make locally generated solutions not match those of the CI system.

Yes. but this one works.  Like the jenkins test https://ci.opm-project.org/job/opm-common-PR-builder/10316/ shows. 

I am okay to undo it and use Jenkins update also. 

This jenkins request in opm-tests just wants to show that the simulator `flow_comp` does not produce the PSAT in the restart file, it confirms.  And that should be the only issue. 

If you prefer, I can remove the reference update and let jenkins do the work.

bska:
> > Did you create those on your own machine? If so, there's a chance that they won't apply to the CI system. 
> 
> Yes. but this one works.

Only until it doesn't.  I just merged #1595 generated by the changes in OPM/opm-common#5348 and that has introduced merge conflicts for the `1D_COMP.*` result files.

> If you prefer, I can remove the reference update and let jenkins do the work.

Yes please. I generally prefer Jenkins generating reference solution updates.

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge it into the master branch in order to prepare for merging the upstream PR.

PR #1597, opened 2026-09-09
Add WPWE summary test cases

GitPaean:


GitPaean:
The cases need to be gone through in more details and testing before deciding whether they are appropriate.

GitPaean:
These are not good cases.

PR #1604, opened 2026-09-11
TPSA case with spring boundary conditions

svenn-t:
Same test case as TPSA_LAGGED, but with a spring boundaries at top and bottom with distance 50 m and 5 GPa shear modulus.


bska (approved):
Thanks.  This looks good to me and I'll merge into master.

